PR #5, opened 2015-07-30
EZP-24649: Solr: Index Locations as nested documents of Content

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-24649
See also: https://github.com/ezsystems/ezpublish-kernel/pull/1358

This implements indexing of Location documents as nested (block joined) documents of Content.
With that change the semantic configuration is simplified, and Location index is no longer separately configured.

### Reasoning

Currently we don't support fulltext search with Location Search. Having Locations as nested documents of Content will enables us implementation of fulltext Location search using block join, for example:

```
{!child of=document_type_id:content v='<fulltext condition>'}
```

That means we don't have to include fulltext field in the Location documents, but can refer to the one in the (parent) Content document. Avoiding that means having better relevancy calculation.

However, since it is not possible to sort and facet using block join, we still must index **all** Content fields in the Location document, same as in the Content document. These will be typically analyzed as keywords and will skew statistics in some measure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ matrix:
         - php: 5.6
           env: TEST_CONFIG="phpunit.xml"
         - php: 5.4
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="4.10.4" CORES_SETUP="dedicated" SOLR_CORES="core0 core1 core2 core3 core4 core5 core6 core7" SOLR_CONFS="lib/eZ/Publish/Core/Search/Solr/Content/Resources/schema.xml"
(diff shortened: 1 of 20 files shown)


andrerom:
+1

pspanja:
Ready for review, ping @andrerom @paulborgermans 

Note the description regarding indexing Content fields and skewing of relevancy, I don't know how much of a problem this can be in the practice and if it is acceptable or not.

bdunogier:
> Note the description regarding indexing Content fields and skewing of relevancy, I don't know how much of a problem this can be in the practice and if it is acceptable or not.

I think it is acceptable, even though I might change my mind depending on the concrete consequences on actual use-cases.

But we must document this as a known issue/thing/feature.

andrerom:
> However, since it is not possible to sort and facet using block join, we still must index all Content fields in the Location document, same as in the Content document. These will be typically analyzed as keywords and will skew statistics in some measure.

But since we can decide to only use it for sorting we can for instance index a subset now right? At least as long as Field SortClause is not supporting custom fields for instance.

andrerom:
+1

pspanja:
Thanks for the reviews!

PR #6, opened 2015-07-30
Implement EZP-24376: Implement language analysis

paulborgermans:
(moved from previous pull request on ezpublish-kernel)


andrerom:
Looks good, but anything remaining besides CS fixes?

paulborgermans:
I have no idea what those CS messages are doing here, it is not about code I committed, isnt there a travi sproblem or so?
take a look at the files I comitted here ... totally different

pspanja:
@paulborgermans I think they were caused by the update to the PHP CS fixer on the CI, as the CS config did not change and it started reporting new problems. I will run the fixer on the whole repo again when the current PRs are merged. I think we can simply ignore it for the moment.

andrerom:
Paul still didn't answer if anything else was remaining here *(besides CS)*.

So assuming it is done, final, ready for review, then: +1

paulborgermans:
Yes, this aspect is done as far as I am concerned :)

andrerom:
@pspanja This one we should also not forget, if it is ready from Paul's side maybe you can have a short look if you can rebase and get it to pass after work on commit is done.

pspanja:
Merged, thanks @paulborgermans

PR #9, opened 2015-08-11
Fix: normal document not overwritten by document block

pspanja:
This fixes a situation where Content document block does not overwrite the existing Content document that is not a block, and vice-versa. The block is a set of nested (block-joined) documents, in our case Content document with one or more nested Location documents.

The problem occurs when Content is first indexed without Locations, and Locations are later added.
In that case Content will first be indexed as a normal (not block-joined) single document, and when Locations are added it will be indexed again, this time as a document block containing Location and Content documents. In this case Solr will use `_root_` field to enforce uniqueness, which means we'll end up with two documents with the same ID. One standalone for Content without Locations (first indexed), and another one inside a document block (indexed later, with Locations).

Reverse situation is also a problem, that is when existing Locations are removed from Content.

The problem is also described here:

* http://grokbase.com/t/lucene/solr-user/14chqr73nv/converting-to-parent-child-block-indexing
* https://issues.apache.org/jira/browse/SOLR-5211

The problem is fixed by enforcing a document block in all cases. This is done by indexing a dummy nested document when Content does not have Locations.

The alternative is sending DELETE request before a document (block) is indexed, in all cases.

### DIFF
--- a/lib/eZ/Publish/Core/Search/Solr/Gateway/Native.php
+++ b/lib/eZ/Publish/Core/Search/Solr/Gateway/Native.php
@@ -413,6 +413,14 @@ class Native extends Gateway
         $xmlWriter->startElement('add');
 
         foreach ($documents as $document) {
+            // Index dummy nested document when there are no other nested documents.
+            // This is done in order to avoid the situation when previous standalone document is
+            // being re-indexed as a block-joined set of documents, or vice-versa.
+            // Enforcing document block in all cases ensures correct overwriting (updating) and
+            // avoiding multiple documents with the same ID.
+            if (empty($document->documents)) {
+                $document->documents[] = $this->getDummyDocument($document->id);
+            }
             $this->writeDocument($xmlWriter, $document);
         }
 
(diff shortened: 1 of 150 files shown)


pspanja:
Ping @andrerom (@paulborgermans)

lib/eZ/Publish/Core/Search/Solr/Gateway/Native.php
```diff
@@ -445,6 +448,22 @@ protected function writeDocument(XmlWriter $xmlWriter, Document $document)
         $xmlWriter->endElement();
     }
 
+    protected function getDummyDocument($id)
```
andrerom: add explanation similar to what you had in the PR so this is understandable by reading the code ;)

andrerom:
+1 with some doc

glye:
Seems reasonable. +0.5 because http://knowyourmeme.com/photos/234137-i-have-no-idea-what-im-doing

pspanja:
Added docs, we should keep an eye on linked Solr issue and remove the workaround when it is fixed on their side.
Not sure what is going on with CS check, seems like it is down rather then failing.

andrerom:
@pspanja you should run it locally, I think @glye have been exposed to how if you don't yet. Casue when there is no comment it just means the comment is to large to be posted so github api refuses.. Does not mean there is to many fixes, just that php-cs-fixer is producing way to big ouput.

pspanja:
Ok, fixed CS issues in a separate commit, merging when Travis finishes as well.

PR #10, opened 2015-08-11
EZP-24704: Solr: Implement native result extractor

pspanja:
> ##### This PR resolves https://jira.ez.no/browse/EZP-24704
> ##### Review together with https://github.com/ezsystems/ezpublish-kernel/pull/1372

This implements *native* result extractor, which extracts search result from data returned by Solr backend.

`NativeResultExtractor` replaces `LoadingResultExtractor`, which did the "extraction" by loading from persistence by ID returned from the Solr backend.

The motive is completely avoiding hitting of persistence when using `SearchService::findContentInfo()`.

### DIFF
--- a/lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
+++ b/lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
@@ -58,7 +58,7 @@ class NativeResultExtractor extends ResultExtractor
                 'name' => $hit->name_s,
                 'contentTypeId' => (int)$hit->type_id,
                 'sectionId' => (int)$hit->section_id,
-                'currentVersionNo' => (int)$hit->version_id,
+                'currentVersionNo' => $hit->version_no_i,
                 'isPublished' => true,
                 'ownerId' => (int)$hit->owner_id,
                 'modificationDate' => strtotime($hit->modified_dt),
--- a/lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
+++ b/lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
@@ -1,0 +1,102 @@
+<?php
+
+/**
+ * This file is part of the eZ Publish Kernel package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
+ */
+namespace eZ\Publish\Core\Search\Solr\ResultExtractor;
+
+use eZ\Publish\Core\Search\Solr\ResultExtractor;
+use eZ\Publish\SPI\Persistence\Content\ContentInfo;
+use eZ\Publish\SPI\Persistence\Content\Location;
+use RuntimeException;
+
+/**
+ * The Native Result Extractor extracts the value object from the data
+ * returned by the Solr backend.
+ */
+class NativeResultExtractor extends ResultExtractor
+{
+    /**
+     * Extracts value object from $hit returned by Solr backend.
+     *
+     * @throws \RuntimeException If search $hit could not be handled
+     *
+     * @param mixed $hit
+     *
+     * @return \eZ\Publish\API\Repository\Values\ValueObject
+     */
+    public function extractHit($hit)
+    {
+        if ($hit->document_type_id === 'content') {
+            return $this->extractContentInfoFromHit($hit);
+        }
+
+        if ($hit->document_type_id === 'location') {
+            return $this->extractLocationFromHit($hit);
+        }
+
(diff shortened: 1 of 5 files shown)


pspanja:
Ready for review, ping @andrerom @glye

lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
```diff
@@ -0,0 +1,102 @@
...
+            "Could not extract: document of type '{$hit->document_type_id}' is not handled."
+        );
+    }
+
+    /**
+     * @param mixed $hit
+     *
+     * @return \eZ\Publish\SPI\Persistence\Content\ContentInfo
+     */
+    protected function extractContentHit($hit)
```
andrerom: `extractContentInfoFromHit`

lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
```diff
@@ -0,0 +1,102 @@
...
+
+        return $contentInfo;
+    }
+
+    /**
+     * @param mixed $hit
+     *
+     * @return \eZ\Publish\SPI\Persistence\Content\Location
+     */
+    protected function extractLocationHit($hit)
```
andrerom: `extractLocationFromHit`

andrerom:
+1

pspanja:
> do we need to cast stuff here maybe? considering the change for weak comparison for kernel test

This implementation should work for any storage, but maybe we can cast for now to avoid breaks on `5.4`. We do hint IDs as `int|string`, so one should not rely on them being `int` alone.

andrerom:
> This implementation should work for any storage, 

ok, ideally id's are either hash or ints, not hash, ints and numerical strings. But unless we index information about this, or abstract data mappers so search engine reuses data mappers from storage engine or something, then I don't see a good way to deal with this cleanly.

lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
pspanja: Updated and rebased.

lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
pspanja: Updated and rebased.

pspanja:
Yeah, you're right. But let's get back at that later :)
Now added casts and rebased, also fixed version number not being indexed as integer in https://github.com/ezsystems/ezplatform-solr-search-engine/commit/7d18cbb7412ee0cffc36408507f251778ffefab4.

lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
```diff
@@ -0,0 +1,102 @@
...
+                'id' => (int)$hit->content_id,
+                'name' => $hit->name_s,
+                'contentTypeId' => (int)$hit->type_id,
+                'sectionId' => (int)$hit->section_id,
+                'currentVersionNo' => $hit->version_no_i,
+                'isPublished' => true,
+                'ownerId' => (int)$hit->owner_id,
+                'modificationDate' => strtotime($hit->modified_dt),
+                'publicationDate' => strtotime($hit->published_dt),
+                'alwaysAvailable' => $hit->always_available_b,
```
andrerom: (bool)?

lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    protected function extractContentInfoFromHit($hit)
+    {
+        $contentInfo = new ContentInfo(
+            array(
+                'id' => (int)$hit->content_id,
+                'name' => $hit->name_s,
+                'contentTypeId' => (int)$hit->type_id,
+                'sectionId' => (int)$hit->section_id,
+                'currentVersionNo' => $hit->version_no_i,
```
andrerom: (int)

andrerom:
Ok, well safer for now.
And as types are probably ok now, +1

lib/eZ/Publish/Core/Search/Solr/ResultExtractor/NativeResultExtractor.php
pspanja: It comes as a bool since it is stored as such and delivered as JSON, which supports PHP's scalar types.
So no conversion needed other than for IDs.

andrerom: ok, great.

PR #11, opened 2015-08-12
Switch to eDisMax query parser

pspanja:
Switching to eDisMax query parser.

eDisMax will provide a base for more potential functionality in the future, for example boost query.


andrerom:
+1

PR #13, opened 2015-08-12
Implement commit() instead of setCommit()

pspanja:
Review together with https://github.com/ezsystems/ezpublish-kernel/pull/1374

This replaces internal search handler method `setCommit()` with `refresh()`, which performs empty soft commit request.

Since indexing is by definition asynchronous/delayed, exposing this in API does not make much sense, since we can't know what exactly will/won't be committed at any given moment.

On the other hand, the method makes sense on the search handler level, which also exposes methods `indexContent()` and `indexLocation()` (these are not async).

### DIFF
--- a/lib/eZ/Publish/API/Repository/Tests/SetupFactory/LegacySolr.php
+++ b/lib/eZ/Publish/API/Repository/Tests/SetupFactory/LegacySolr.php
@@ -130,8 +130,8 @@ class LegacySolr extends Legacy
 
         /** @var \eZ\Publish\Core\Search\Solr\Handler $searchHandler */
         $searchHandler->purgeIndex();
-        $searchHandler->setCommit(true);
         $searchHandler->bulkIndexContent($contentObjects);
+        $searchHandler->commit();
     }
 
     protected function getTestConfigurationFile()
--- a/lib/eZ/Publish/Core/Search/Solr/Gateway.php
+++ b/lib/eZ/Publish/Core/Search/Solr/Gateway.php
@@ -63,11 +63,13 @@ abstract class Gateway
     abstract public function purgeIndex();
 
     /**
-     * Set if index/delete actions should commit or if several actions is to be expected.
+     * Commits the data to the Solr index, making it available for search.
      *
-     * This should be set to false before group of actions and true before the last one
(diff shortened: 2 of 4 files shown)


pspanja:
Ping @andrerom @glye

lib/eZ/Publish/API/Repository/Tests/SetupFactory/LegacySolr.php
```diff
@@ -130,8 +130,8 @@ protected function indexAll()
 
         /** @var \eZ\Publish\Core\Search\Solr\Handler $searchHandler */
         $searchHandler->purgeIndex();
-        $searchHandler->setCommit(true);
         $searchHandler->bulkIndexContent($contentObjects);
+        $searchHandler->refresh();
```
andrerom: hmm, not a huge deal, but thought we said `->commit($flush = false);`
For me at least, flush, and especially refresh does not speak to me, but in the context of commit they do, then I know the context.

pspanja: I can change that, I went for this thinking it would be good to make a distinction between DB commit, and this is something that would be typically called outside of transaction. (TBH that was before I concluded it makes no sense to expose this in API)

pspanja: Renamed to `commit()` and added `$flush` parameter in https://github.com/ezsystems/ezplatform-solr-search-engine/commit/3c21712ee22eefc5487998c891605f4d6fb38a06

andrerom:
+1

pspanja:
The job seems to be running, it is like only the wrong page is loaded on Travis (This is not an active repository). You can even see the duration of running when searching for the repo in the left side menu.
Maybe someone with admin rights there can click on some switches?

Anyways I'm just merging this and the other one when it finishes.

andrerom:
the thing is clicking on "settings" give the same screen, so no idea.

pspanja:
Looks like it just decided it will work :) Something you did or?

pspanja:
Ok, thanks :)

PR #14, opened 2015-08-19
EZP-24718: Remove deprecated criteria and sort clauses

pspanja:
> ##### This PR resolves https://jira.ez.no/browse/EZP-24718
> ##### Review together with https://github.com/ezsystems/ezpublish-kernel/pull/1378

This removes visitors for deprecated criteria and sort clauses:

* `eZ\Publish\API\Repository\Values\Content\Query\Criterion\Depth`
* `eZ\Publish\API\Repository\Values\Content\Query\Criterion\LocationPriority`
* `eZ\Publish\API\Repository\Values\Content\Query\SortClause\LocationDepth`
* `eZ\Publish\API\Repository\Values\Content\Query\SortClause\LocationPath`
* `eZ\Publish\API\Repository\Values\Content\Query\SortClause\LocationPathString`
* `eZ\Publish\API\Repository\Values\Content\Query\SortClause\LocationPriority`


pspanja:
Ready for review, ping @andrerom

andrerom:
Is this something we want to do yet considering we want to support 5.4 in 1.0?

pspanja:
My idea is we remove support for those on 5.4 with Solr, but not with Legacy engine, so ideally we run the same engine version both on 5.4 and 1.0. Would you rather keep support on 5.4 with Solr?

andrerom:
Your argument make sense, as in people having to avoid using deprecated features to use Solr.

So +1 as long as legacy search engine on 5.4 works like before.

pspanja:
> So +1 as long as legacy search engine on 5.4 works like before.

Ok, that is the idea.

gggeek:
Q: how could someone figure out, when on 5.4, if the current search engine dos support these sort/filter criteria?
So far in KMB I did use a class_exist check, which works with recent ez-kernel packages. However, with this installed, it becomes hard to decide if a query can be created with those criteria, as they do exist...

pspanja:
@gggeek I think checking for the search handler instance and existence of visitors for these criteria should work.

pspanja:
BTW, if using these criteria you might find this useful, it should work on 5.4 as well: https://docs.netgen.io/projects/search-extra/en/latest/reference/location_query.html

gggeek:
@pspanja thanks for the quick answer.
 
I started taking a look at the existing services... ideally I would be able to start from `$repository->getSearchService()` and drill down all the way to the list of registered visitors, but of course dependencies are all protected or private, so this is impossible (unless using the `apply` hack to access protected class members).

Which leaves me with the option of checking the registered visitors in the `ezpublish.search.solr.query.common.sort_clause_visitor.aggregate` service by calling `canVisit()`, and hope that the end user has not overridden any of the injected services through the chain.

Either that, or I try the query in the 1st place and check if it throws a NotImplementedException...

ps: thx for the location_query suggestion - but at the moment I am only interested in the Sort clause.

pspanja:
I'm probably missing something, but if the visitor exists it should be registered, so why not just using `class_exists` on it? If needed you can also combine that with checking on the instance of the search handler.

gggeek:
Unfortunately the class does exist, as it comes from ez 5.4, which still has it...

ps: moving the discussion to Slack, as it is getting a bit long for a comment on an old commit.

PR #16, opened 2015-08-24
Adjust ezsystems/ezpublish-kernel dependency

pspanja:
This makes possible to install the engine with latest ezpublish-kernel `master` and `5.4` stable.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "~4.7",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -8,13 +8,14 @@
             "email": "dev-team@ez.no"
         }
     ],
-    "minimum-stability": "stable",
+    "minimum-stability": "alpha",
```
andrerom: Just for reference this is temporary until we have stable tags of 5.4.5 and 6.0.0.

PR #18, opened 2015-09-15
Implement EZP-24802: Full text Location search

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-24802

This implements full text Location search using child block join, possible since changing indexing of Locations to subdocuments of Content in https://jira.ez.no/browse/EZP-24649. The idea behind this was reusing parent document (Content) relevancy with full text Location search.

There is a bug in Solr version we currently support (`4.10`), causing skipping of scoring when using child block join: https://issues.apache.org/jira/browse/LUCENE-6588.
The issue is fixed since Solr `5.3` (latest).

### DIFF
--- a/lib/eZ/Publish/Core/Search/Solr/Query/Content/CriterionVisitor/FullText.php
+++ b/lib/eZ/Publish/Core/Search/Solr/Query/Content/CriterionVisitor/FullText.php
@@ -83,8 +83,7 @@ class FullText extends CriterionVisitor
             }
         }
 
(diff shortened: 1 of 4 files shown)


pspanja:
Ready for review, ping @andrerom @paulborgermans

andrerom:
+1

pspanja:
Merging this to unblock other stuff, @paulborgermans if you have any remarks feel free to do it regardless.

PR #19, opened 2015-09-15
Implement EZP-24804: Ancestor criterion

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-24804
Review together with https://github.com/ezsystems/ezpublish-kernel/pull/1409

This implements `Ancestor` criterion visitors, accepting Location path string(s) as value.


andrerom:
in support seems to be missing, should we skip it form api if we don't intend to impl it anytime soon?

pspanja:
`IN` is already handled in the same visitor, but `canVisit()` was allowing `EQ` only.
Now updated and rebased.

andrerom:
+1

lolautruche:
+1

PR #20, opened 2015-09-16
Implement EZP-24816: Indexable for Url field type

pspanja:
#### This PR https://jira.ez.no/browse/EZP-24816

This configures searchable Url field type to enable tests.

### DIFF
--- a/lib/eZ/Publish/Core/settings/tests/solr/single_core.yml
+++ b/lib/eZ/Publish/Core/settings/tests/solr/single_core.yml
@@ -1,16 +1,11 @@
 parameters:
-    # Redefining default ezfloat field type class with the implementation that defines the type
+    # Redefining default field type classes with the implementation that defines the type
     # as searchable. Needed in order to test searching wiht Solr engine.
     ezpublish.fieldType.ezfloat.class: eZ\Publish\API\Repository\Tests\FieldType\SearchableFloat
-    # Redefining default ezimage field type class with the implementation that defines the type
-    # as searchable. Needed in order to test searching wiht Solr engine.
     ezpublish.fieldType.ezimage.class: eZ\Publish\API\Repository\Tests\FieldType\SearchableImage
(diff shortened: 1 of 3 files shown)


lib/eZ/Publish/Core/settings/tests/solr/single_core.yml
```diff
@@ -1,16 +1,11 @@
...
-    # Redefining default ezimage field type class with the implementation that defines the type
-    # as searchable. Needed in order to test searching wiht Solr engine.
     ezpublish.fieldType.ezimage.class: eZ\Publish\API\Repository\Tests\FieldType\SearchableImage
-    # Redefining default eztext field type class with the implementation that defines the type
-    # as searchable. Needed in order to test searching wiht Solr engine.
     ezpublish.fieldType.eztext.class: eZ\Publish\API\Repository\Tests\FieldType\SearchableTextBlock
-    # Redefining default ezmedia field type class with the implementation that defines the type
-    # as searchable. Needed in order to test searching wiht Solr engine.
     ezpublish.fieldType.ezmedia.class: eZ\Publish\API\Repository\Tests\FieldType\SearchableMedia
+    ezpublish.fieldType.ezurl.class: eZ\Publish\API\Repository\Tests\FieldType\SearchableUrl
```
andrerom: This is rather strange, isn't it more in line with other types to mark it as searchable and throw Not Implemented for Legacy?

andrerom: ah, this is only for tests. how is it searchable in normal use?

pspanja: Yes :) It isn't searchable, we should decide what to do about it.
Providing searchable versions like these to configure as needed should be ok I think.

pspanja: One other option is to define them in the bundle, so they would be searchable if the engine is enabled.

PR #21, opened 2015-09-21
Implement EZP-23247: missing field type Indexable definitions

pspanja:
> #### This PR resolves https://jira.ez.no/browse/EZP-23247
> #### Review with https://github.com/ezsystems/ezpublish-kernel/pull/1418

This implements handling of the new SPI search backend field type `FullText`.
Fields of this type are not stored, but instead are copied to the real full text field `text`.

This is used by the field type's `Indexable` definitions for indexing full text data.

#### Additionaly fixed

* removed handling of `DOMDocument` from string field value mapper (cleanup after merge of https://github.com/ezsystems/ezpublish-kernel/pull/262)
* referencing non-existent `HtmlField` SPI search backend field type

### DIFF
--- a/lib/eZ/Publish/Core/Search/Solr/DocumentMapper/NativeDocumentMapper.php
+++ b/lib/eZ/Publish/Core/Search/Solr/DocumentMapper/NativeDocumentMapper.php
@@ -323,7 +323,7 @@ class NativeDocumentMapper implements DocumentMapper
             new FieldType\MultipleIdentifierField()
         );
 
-        $fieldSets = $this->mapContentFields($content, $contentType, true);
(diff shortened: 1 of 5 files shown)


andrerom:
I guess this is somewhat wip, but review ping @paulborgermans

pspanja:
@andrerom This is ready actually, together with https://github.com/ezsystems/ezpublish-kernel/pull/1418.

lolautruche:
+1 with tests fixed

pspanja:
@lolautruche fails because of fixed kernel version, will try to fix that today.

andrerom:
+1

PR #23, opened 2015-10-06
Fix EZP-24902: Document is not indexed for translation

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-24902
Review together with https://github.com/ezsystems/ezpublish-kernel/pull/1441

When mapping Content to documents per translation, translations are determined from the generated document fields. Since field is not mapped to the document if value of the field equals to `null`, it can happen that document is not added for the translation if all field values for the translation equal to `null`.

With this change a document will be indexed for translation even when it has no Content fields.

Additionally fixed indexing full-text fields to Location documents.


pspanja:
Review ping @andrerom @bdunogier @glye @lolautruche @yannickroger

lolautruche:
You don't like short array syntax, do you? :smile: 

+1

glye:
+1

pspanja:
I don't particularly dislike it, mostly I don't remember it exists or just want to keep things uniform :)
Thanks for the review!

PR #24, opened 2015-10-08
Fix EZP-24915: Solr Search Engine repository structure cleanup

pspanja:
#### This PR resolves https://jira.ez.no/browse/EZP-24915

Corresponding PRs:
* https://github.com/ezsystems/ezpublish-kernel/pull/1452
* https://github.com/ezsystems/ezpublish-platform/pull/18
* https://github.com/ezsystems/ezplatform/pull/53
* https://github.com/ezsystems/ezpublish-community/pull/250

This mostly reduces the depth of directory structure:

*  `bundle/EzPublishSolrSearchEngineBundle` → `bundle`
* `lib/eZ/Publish/Core/...` → `lib/...`

  The namespace of implementation was also changed, now the implementation is under `EzSystems\EzPlatformSolrSearchEngine` namespace.
* integration tests setup factory is moved to `tests/lib` directory

Additionally standard bundle namespace is adopted:

`eZ\Bundle\EzPublishSolrSearchEngineBundle` → `EzSystems\EzPlatformSolrSearchEngineBundle`


pspanja:
Review ping @andrerom @bdunogier @lolautruche

andrerom:
Still deep:
> Base/Container/Compiler/Search/Solr

Any reason why we use the word "settings" when "config" tends to be the Symfony norm?

pspanja:
> Still deep:

The idea was to keep the `eZ/Publish/Core` as common root, so that it is easier to orient when coming from kernel, as this does implement Repository API and SPI.

> Any reason why we use the word "settings" when "config" tends to be the Symfony norm?

That part is not Symfony ;)
I kept the naming the same as in kernel, for the same reason as above.

pspanja:
BTW, note that namespaces are separated in autoload, that should have better performance.

andrerom:
> BTW, note that namespaces are separated in autoload, that should have better performance.

what do you mean?

pspanja:
I meant having this structure under `lib/` should not have negative impact on autoloading :)

andrerom:
true :)

But I was more interested in not recreating kernel structure as it's structure is not necessarily a good fit for bundles and libs, so should not dictate how we structure them :)

pspanja:
Then how about:

1. `lib/Base/Container/Compiler/Search/Solr/*` → `lib/Container/Compiler/*`
2. `lib/Search/Solr/*` → `lib/*`
3. `lib/settings/*` → `lib/config/*`

I would rather we agree on this now, so I can apply the same on XmlText repo.

andrerom:
> lib/Base/Container/Compiler/Search/Solr/* → lib/Container/Compiler/*
yes

> lib/Search/Solr/* → lib/*

n/a, I didn't have anything against that one

> lib/settings/* → lib/Resources/config/*

n/a, this is the lib and not the bundle, so would need input from @bdunogier and @lolautruche on that.

andrerom:
>> lib/Search/Solr/* → lib/*
> 
> n/a, I didn't have anything against that one

but then again, this is Solr search bundle, so indeed it could be skipped.

andrerom:
>> lib/settings/* → lib/Resources/config/*
> 
> n/a, this is the lib and not the bundle, so would need input from @bdunogier and @lolautruche on that.

With lack of feedback lets go for `lib/config/`, there is no public folder/css/js planned in the lib.

pspanja:
While we are at it, how about changing the namespace as well:

`eZ\Publish\Core\Search\Solr` → `EzSystems\SolrSearchEngine`

pspanja:
Updated in https://github.com/ezsystems/ezplatform-solr-search-engine/commit/d664195a06305d39bd2fe310eb666e42bc95c024

pspanja:
FWIW I changed the namespace in https://github.com/ezsystems/ezplatform-solr-search-engine/commit/cda636f8d01f3594edff775f73fe6b257ae077bb

andrerom:
> While we are at it, how about changing the namespace as well:
> eZ\Publish\Core\Search\Solr → EzSystems\SolrSearchEngine

yes :)

composer.json
```diff
@@ -19,12 +19,11 @@
...
         "psr-4": {
-            "eZ\\Bundle\\EzPublishSolrSearchEngineBundle\\": "bundle/EzPublishSolrSearchEngineBundle/",
-            "eZ\\Publish\\API\\Repository\\Tests\\SetupFactory\\": "lib/eZ/Publish/API/Repository/Tests/SetupFactory/",
-            "eZ\\Publish\\Core\\Search\\Solr\\": "lib/eZ/Publish/Core/Search/Solr/",
-            "eZ\\Publish\\Core\\Base\\Container\\Compiler\\Search\\Solr\\": "lib/eZ/Publish/Core/Base/Container/Compiler/Search/Solr",
-            "eZ\\SolrSearchEngineBundle\\Tests\\": "tests/bundle",
-            "eZ\\SolrSearchEngine\\Tests\\": "tests/lib"
+            "EzSystems\\SolrSearchEngine\\": "lib",
+            "EzSystems\\SolrSearchEngine\\Tests\\": "tests/lib",
+            "eZ\\Publish\\API\\Repository\\Tests\\": "tests/integration",
```
andrerom: these two, and the last one below can be moved into autoload-dev

as for the namespace, it kind of feels wrong, wdyt?

andrerom:
>>> lib/settings/* → lib/Resources/config/*
>> n/a, this is the lib and not the bundle, so would need input from @bdunogier and @lolautruche on that.
> With lack of feedback lets go for lib/config/, there is no public folder/css/js planned in the lib.

Just noticed you already have a Resources folder for sorl config, so ignore what I said and move the config folder into there like you proposed :)

bundle/ApiLoader/SolrEngineFactory.php
```diff
@@ -8,7 +8,7 @@
  *
  * @version //autogentag//
  */
-namespace eZ\Bundle\EzPublishSolrSearchEngineBundle\ApiLoader;
+namespace EzSystems\SolrSearchEngineBundle\ApiLoader;
```
andrerom: Sorry.. I know this affects all, but given this is not a bundle on it's own, it kind of has to be:
`EzSystems\EzPlatformSolrSearchEngineBundle`

So sorry for going a bit back and forward on this, but this is template for many other future bundles :)

pspanja: No worries, I'm on this as soon as I dig myself out of the XmlText a little bit :)

pspanja: BTW, shouldn't we then go for `EzSystems\EzPlatformSolrSearchEngine` as well?

andrerom: context? where? for what? :)

pspanja: For lib stuff, yes :)

andrerom: yes indeed

composer.json
pspanja: Yes, I will change that as well.

pspanja:
Updated bundle namespace in https://github.com/ezsystems/ezplatform-solr-search-engine/commit/d0e26041d4ed005a7e94da733e7efaf4087beba4.

pspanja:
Updated implementation namespace in https://github.com/ezsystems/ezplatform-solr-search-engine/commit/a90ff5e5d80fa14dabe005243fdde8019a3fa5ab.

pspanja:
@andrerom Updated tests namespace in https://github.com/ezsystems/ezplatform-solr-search-engine/commit/8aed5f9ee2b71a52a6e94204511cde022ca17269 and moved configuration around in https://github.com/ezsystems/ezplatform-solr-search-engine/commit/921ef5d9729fdcd6dba71fb0ac159a3735b108f5.

Container config is now in `lib/Resources/config/container/...`, Solr configuration files are moved to `lib/Resources/config/solr/...`.
Test configuration is moved to `tests/lib/Resources/config/...`.

Setup factory is moved to `tests/lib/SetupFactory/LegacySetupFactory`.

This will now fail while the kernel is not updated, but it is complete on this side and can be reviewed.

andrerom:
Ok, that is good enough :)

+1

bdunogier:
+1

PR #25, opened 2015-10-16
Updated file docblocks

pspanja:
This just updates the first line in PHP file docblocks to:

`This file is part of the eZ Platform Solr Search Engine package.`


pspanja:
BTW this will now fail (master included) until new version of kernel is tagged.

andrerom:
+1 _(will need to update doc for when we tag)_

PR #26, opened 2015-10-26
Fixed EZP-24915: updated endpoint class constant

pspanja:
Endpoint class in bundle extension was forgotten in the restructuring.


bdunogier:
+1

PR #27, opened 2015-11-07
Check for configured engine in the indexing command

pspanja:
Registering Solr Search Engine bundle but forgetting configuration of the repository with `solr` search engine is likely to be one of the common mistakes, resulting in cryptic fatal error.

This updates indexing command with check for configured engine, throwing an exception with user friendly message if expected engine is not found.


pspanja:
Ping @andrerom @bdunogier

andrerom:
+1

bdunogier:
+1

PR #30, opened 2015-12-03
Fix EZP-25187: float/double indexing can fail due to locale

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25187

Different locales can represet floats differently (such as `123,45` for `de_DE.UTF-8`), which cause solr indexing to fail.
This uses the same, simple solution adopted for ezfind's indexboost (https://github.com/ezsystems/ezfind/commit/f46a315f2fa117e7).

### DIFF
--- a/lib/FieldValueMapper/FloatMapper.php
+++ b/lib/FieldValueMapper/FloatMapper.php
@@ -40,6 +40,19 @@ class FloatMapper extends FieldValueMapper
      */
     public function map(Field $field)
     {
-        return (float)$field->value;
+        return $this->fixupFloat($field->value);
+    }
+
+    /**
+     * Convert to a proper Solr representation.
+     *
+     * @param mixed $value
+     *
+     * @return string
+     */
+    protected function fixupFloat($value)
+    {
+        // This will force the '.' as decimal separator and not depend on the locale
+        return sprintf('%F', (float)$value);
     }
 }
--- a/lib/Query/CriterionVisitor.php
+++ b/lib/Query/CriterionVisitor.php
@@ -108,6 +108,8 @@ abstract class CriterionVisitor
         switch (gettype($value)) {
             case 'boolean':
                 return ($value ? 'true' : 'false');
+            case 'double':
+                return sprintf('%F', $value);
 
             default:
                 return (string)$value;


andrerom:
+1

joaoinacio:
Good point, haven't thought about that.
Actually, looking `CriterionVisitor::toString()` it seems the same conversion may be needed
(ref:  https://github.com/ezsystems/ezplatform-solr-search-engine/blob/master/lib/Query/CriterionVisitor.php#L106 )

I'll do a couple of tests and update as necessary, thanks for the feedback!

lib/FieldValueMapper/FloatMapper.php
```diff
@@ -40,6 +40,19 @@ public function canMap(Field $field)
...
+    }
+
+    /**
+     * Convert to a proper Solr representation.
+     *
+     * @param mixed $value
+     *
+     * @return string
+     */
+    protected function convert($value)
```
bdunogier: That's a bit generic. It could be converting it to a baked potato, for all we know :-)

`fixupFloat` would work (even though we could skip the protected method. The only benefit is the method's doc, so I'm fine with it.

bdunogier:
+1 besides the naming nitpick.

lib/FieldValueMapper/FloatMapper.php
joaoinacio: I just copy-pasted it from https://github.com/joaoinacio/ezplatform-solr-search-engine/blob/EZP-25187_float_locale/lib/FieldValueMapper/StringMapper.php#L50 but any name works for me.

joaoinacio:
Updated.
Conversion also added to criterion visitor, even though float fieldtypes aren't really searchable without hacking the database (then it works fine).

lib/FieldValueMapper/FloatMapper.php
```diff
@@ -40,6 +40,19 @@ public function canMap(Field $field)
      */
     public function map(Field $field)
     {
-        return (float)$field->value;
+        return $this->fixupFloat((float)$field->value);
```
bdunogier: Actually, why not pass the value without a cast, and do the cast in the sprintf call ?

bdunogier:
Added a nitpick about casting, but +1.

lib/FieldValueMapper/FloatMapper.php
```diff
@@ -40,6 +40,19 @@ public function canMap(Field $field)
...
+    }
+
+    /**
+     * Convert to a proper Solr representation.
+     *
+     * @param mixed $value
+     *
+     * @return string
+     */
+    protected function fixupFloat($value)
```
andrerom: as this is not reused, and it's one line we don't strictly need a function for this.

andrerom:
+1

PR #31, opened 2015-12-04
Fix EZP-25158: Missing FieldRelation Criterion visitor

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25158

Attempting to use the `FieldRelation` Criterion will result in a "Intentionally not implemented" exception.

As the relation fields are already being indexed, the approach is to simply reuse/extend the Field criterion.

### DIFF
--- a/lib/Resources/config/container/solr/criterion_visitors.yml
+++ b/lib/Resources/config/container/solr/criterion_visitors.yml
@@ -28,6 +28,7 @@ parameters:
     ezpublish.search.solr.query.content.criterion_visitor.map_location_distance_in.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\MapLocation\MapLocationDistanceIn
     ezpublish.search.solr.query.content.criterion_visitor.map_location_distance_range.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\MapLocation\MapLocationDistanceRange
     ezpublish.search.solr.query.content.criterion_visitor.field_in.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\Field\FieldIn
+    ezpublish.search.solr.query.content.criterion_visitor.field_relation.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\Field\FieldRelation
     ezpublish.search.solr.query.content.criterion_visitor.field_range.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\Field\FieldRange
     ezpublish.search.solr.query.content.criterion_visitor.visibility.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\Visibility
     ezpublish.search.solr.query.content.criterion_visitor.custom_field_in.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\CustomField\CustomFieldIn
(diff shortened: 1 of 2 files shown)


bdunogier:
Looks okay. I'm assuming that the relation list is indexed as an... array of content id ?

What does the Criteria look like in that case, if I may ask ?

That aside, +1 if it works. No way to add coverage for this (not sure how extensive the tests are) ?

joaoinacio:
@bdunogier : 
1. yes:
...
        "test_class_relation_value_s": "179",
        "test_class_relations_value_ms": [
          "179"
        ],
```
2. Do you mean the solr query?
```
(location_visible_b:true+AND+test_article_relations_value_ms:"179"+OR+test_class_relations_value_ms:"179"+OR+test_article_relations_value_ms:"123"+OR+test_class_relations_value_ms:"123")+AND+(type_id:"42"))
```
3. Seems to, both for relation and relations. I will check about tests.

**Comment by joaoinacio** [comment]
Regarding tests, there is only a `FullText` Criterion visitor test:
https://github.com/ezsystems/ezplatform-solr-search-engine/tree/master/tests/lib/Search/Query/CriterionVisitor

Thoughts?

**Comment by andrerom** [comment]
@joaoinacio I as referring to Integration tests.

**Comment by andrerom** [comment]
added, see link above, if it passes I'll merge and re run the tests here.

**Comment by andrerom** [comment]
@joaoinacio seems to fail on all IN cases, and tests are most likely skipping on all Operator::CONTAINS cases.

**Comment by andrerom** [comment]
@joaoinacio I don't think you can use Field as is, the FieldRelation does not seem to have contentType identifier as expected input for instance.

**Comment by joaoinacio** [comment]
@andrerom: thanks, I'll check on that and the `CONTAINS`

Meanwhile, it seems there is a problem with the tests, AFAICT a `$this->refreshSearch($repository);` call is missing after content publication (causing empty results leading to fails).

**Comment by andrerom** [comment]
Merged the test fix and restarted test

File: lib/Query/Content/CriterionVisitor/Field/FieldRelationIn.php [hunk 1]
```diff
@@ -0,0 +1,81 @@
...
+
+use EzSystems\EzPlatformSolrSearchEngine\Query\CriterionVisitor;
+use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
+use eZ\Publish\API\Repository\Values\Content\Query\Criterion\Operator;
+use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
+
+/**
+ * Visits the FieldRelation criterion.
+ */
+class FieldRelationIn extends FieldIn
```
andrerom: Can skip "In" in name now

lib/Resources/config/container/solr/criterion_visitors.yml
```diff
@@ -28,6 +28,7 @@ parameters:
     ezpublish.search.solr.query.content.criterion_visitor.map_location_distance_in.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\MapLocation\MapLocationDistanceIn
     ezpublish.search.solr.query.content.criterion_visitor.map_location_distance_range.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\MapLocation\MapLocationDistanceRange
     ezpublish.search.solr.query.content.criterion_visitor.field_in.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\Field\FieldIn
+    ezpublish.search.solr.query.content.criterion_visitor.field_relation_in.class: EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\Field\FieldRelationIn
```
andrerom: same

lib/Resources/config/container/solr/criterion_visitors.yml
```diff
@@ -225,6 +226,13 @@ services:
         tags:
             - {name: ezpublish.search.solr.query.content.criterion_visitor}
 
+    ezpublish.search.solr.query.content.criterion_visitor.field_relation_in:
```
andrerom: same

andrerom:
Besides nitpics, +1

andrerom:
Thanks @joaoinacio, will tag a 1.0.2-rc1 with this so user can try it.

PR #32, opened 2015-12-10
[Doc] Add info in Readme for running tests

andrerom:


joaoinacio:
Nitpick: content seems fine, but list numbering is restarted (1.  2.  1.  1. ) when viewed through github

PR #33, opened 2015-12-11
Fix EZP-25246 MapLocation issues with alternate locale

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25246

This fixes the exception caused by wrong float format when indexing content, also when searching.


andrerom:
+1 (with cs fixes)

PR #34, opened 2015-12-23
Fix EZP-25334: Missing PermissionSubtree Criterion Visitor

harmstyler:
PermissionVisitor was not implemented yet, this causes an uncaught exception when controlling user permissions by subtree.

### DIFF
--- a/lib/Query/Location/CriterionVisitor/PermissionSubtree.php
+++ b/lib/Query/Location/CriterionVisitor/PermissionSubtree.php
@@ -33,6 +33,7 @@ class PermissionSubtree extends CriterionVisitor
     public function visit(Criterion $criterion, CriterionVisitor $subVisitor = null)
     {
         $subtrees = $criterion->value;
+
         return 'path_string_id:[' . implode('* ', $subtrees) . '*]';
     }
 }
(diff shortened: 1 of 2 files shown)


andrerom:
I guess you just copied SubtreeVisitor here?

I would rather fix this in Core actually *(Making PermissionSubtree extend Subtree, and making sure Legacy Engine, which has this issue, picks the enhanced one)*.

harmstyler:
I'm not sure what `SubtreeVisitor` is, maybe `SubtreeIn`? My PR just adds another visitor for when `$criterion` of class `eZ\Publish\Core\Repository\Values\Content\Query\Criterion\PermissionSubtree` is passed.

It would make sense to fix in core if Core and reuse what's already there. Any chance that fix would make it into the 5.4.x releases?

andrerom:
> It would make sense to fix in core and reuse what's already there. Any chance that fix would make it into the 5.4.x releases?

sure, but we

pspanja:
I think this can be best fixed on the Legacy Search Engine side, as @andrerom suggested.
See https://github.com/ezsystems/ezpublish-kernel/pull/912#discussion_r49459938.

andrerom:
PR created, review and testing more then welcome :)
 https://github.com/ezsystems/ezpublish-kernel/pull/1562

PR #35, opened 2015-12-28
Fix EZP-25309: invalid date string using field criterion

joaoinacio:
> JIRA: https://jira.ez.no/browse/EZP-25309
> ~~Depends on https://github.com/ezsystems/ezpublish-kernel/pull/1551~~

This updates the Field criterion visitor to apply custom "filtering" based on the target field type, resolving the problem of having timestamps for querying solr datetimes.

TODO:
- [x] fix unit tests failure(s)

### DIFF
--- a/lib/Query/Content/CriterionVisitor/Field.php
+++ b/lib/Query/Content/CriterionVisitor/Field.php
@@ -11,8 +11,11 @@
 namespace EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor;
 
 use EzSystems\EzPlatformSolrSearchEngine\Query\CriterionVisitor;
+use EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper;
 use eZ\Publish\Core\Search\Common\FieldNameResolver;
 use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
+use eZ\Publish\SPI\Search\Field as SearchField;
+use eZ\Publish\SPI\FieldType\Indexable;
 
 /**
  * Visits the Field criterion.
@@ -27,36 +30,60 @@ abstract class Field extends CriterionVisitor
     protected $fieldNameResolver;
 
     /**
+     * @var \EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper
+     */
+    protected $fieldValueMapper;
+
+    /**
      * Create from content type handler and field registry.
      *
      * @param \eZ\Publish\Core\Search\Common\FieldNameResolver $fieldNameResolver
+     * @param \EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper $fieldValueMapper
      */
-    public function __construct(FieldNameResolver $fieldNameResolver)
+    public function __construct(FieldNameResolver $fieldNameResolver, FieldValueMapper $fieldValueMapper)
     {
         $this->fieldNameResolver = $fieldNameResolver;
+        $this->fieldValueMapper = $fieldValueMapper;
     }
 
     /**
-     * Get field names.
+     * Get array of search fields.
      *
      * @param \eZ\Publish\API\Repository\Values\Content\Query\Criterion $criterion
-     * @param string $fieldDefinitionIdentifier
-     * @param string $fieldTypeIdentifier
-     * @param string $name
      *
-     * @return array
+     * @return array array of field types indexed by name.
      */
-    protected function getFieldNames(
-        Criterion $criterion,
-        $fieldDefinitionIdentifier,
-        $fieldTypeIdentifier = null,
-        $name = null
-    ) {
-        return $this->fieldNameResolver->getFieldNames(
+    protected function getSearchFields(Criterion $criterion)
+    {
+        return $this->fieldNameResolver->getFieldTypes(
             $criterion,
-            $fieldDefinitionIdentifier,
-            $fieldTypeIdentifier,
-            $name
+            $criterion->target
         );
+    }
+
+    /**
+     * Map search field value to solr value using FieldValueMapper.
+     *
+     * @param eZ\Publish\SPI\FieldType\Indexable $searchFieldType
+     * @param mixed $value
+     *
+     * @return mixed
+     */
+    protected function mapSearchFieldValue(Indexable $searchFieldType, $value)
+    {
+        $definition = $searchFieldType->getIndexDefinition();
+        if ($searchFieldType instanceof \eZ\Publish\Core\FieldType\Selection\SearchField) {
+            $matchField = 'selected_option_value';
+        } else {
+            $matchField = $searchFieldType->getDefaultMatchField();
+        }
+
+        $fieldType = $definition[$matchField];
+        // convert core fieldtype search into a SPI/Search/Field to use in FieldValueMapper
+        $searchField = new SearchField('field', $value, $fieldType);
+
+        $value = $this->fieldValueMapper->map($searchField);
+
+        return current((array)$value);
     }
 }
--- a/lib/Query/Content/CriterionVisitor/Field/FieldIn.php
+++ b/lib/Query/Content/CriterionVisitor/Field/FieldIn.php
@@ -49,9 +49,9 @@ class FieldIn extends Field
      */
     public function visit(Criterion $criterion, CriterionVisitor $subVisitor = null)
     {
-        $fieldNames = $this->getFieldNames($criterion, $criterion->target);
+        $searchFields = $this->getSearchFields($criterion);
 
-        if (empty($fieldNames)) {
+        if (empty($searchFields)) {
             throw new InvalidArgumentException(
                 '$criterion->target',
                 "No searchable fields found for the given criterion target '{$criterion->target}'."
@@ -59,12 +59,12 @@ class FieldIn extends Field
         }
 
         $criterion->value = (array)$criterion->value;
-
         $queries = array();
-        foreach ($criterion->value as $value) {
-            $preparedValue = $this->escapeQuote($this->toString($value), true);
 
-            foreach ($fieldNames as $name) {
+        foreach ($searchFields as $name => $fieldType) {
+            foreach ($criterion->value as $value) {
+                $preparedValue = $this->escapeQuote($this->mapSearchFieldValue($fieldType, $value), true);
+
                 $queries[] = $name . ':"' . $preparedValue . '"';
             }
         }
--- a/lib/Query/Content/CriterionVisitor/Field/FieldRange.php
+++ b/lib/Query/Content/CriterionVisitor/Field/FieldRange.php
@@ -51,27 +51,27 @@ class FieldRange extends Field
      */
     public function visit(Criterion $criterion, CriterionVisitor $subVisitor = null)
     {
-        $value = (array)$criterion->value;
-        $start = $value[0];
-        $end = isset($value[1]) ? $value[1] : null;
+        $searchFields = $this->getSearchFields($criterion);
 
-        if (($criterion->operator === Operator::LT) ||
-             ($criterion->operator === Operator::LTE)) {
-            $end = $start;
-            $start = null;
-        }
-
-        $fieldNames = $this->getFieldNames($criterion, $criterion->target);
-
-        if (empty($fieldNames)) {
+        if (empty($searchFields)) {
             throw new InvalidArgumentException(
                 '$criterion->target',
                 "No searchable fields found for the given criterion target '{$criterion->target}'."
             );
         }
 
+        $value = (array)$criterion->value;
         $queries = array();
-        foreach ($fieldNames as $name) {
+        foreach ($searchFields as $name => $fieldType) {
+            $start = $this->mapSearchFieldValue($fieldType, $value[0]);
+            $end = isset($value[1]) ? $this->mapSearchFieldvalue($fieldType, $value[1]) : null;
+
+            if (($criterion->operator === Operator::LT) ||
+                  ($criterion->operator === Operator::LTE)) {
+                $end = $start;
+                $start = null;
+            }
+
             $queries[] = $name . ':' . $this->getRange($criterion->operator, $start, $end);
         }
 
--- a/lib/Query/Content/CriterionVisitor/Field/FieldRelation.php
+++ b/lib/Query/Content/CriterionVisitor/Field/FieldRelation.php
@@ -48,9 +48,9 @@ class FieldRelation extends Field
      */
     public function visit(Criterion $criterion, CriterionVisitor $subVisitor = null)
     {
-        $fieldNames = $this->getFieldNames($criterion, $criterion->target);
+        $searchFields = $this->getSearchFields($criterion, $criterion->target);
 
-        if (empty($fieldNames)) {
+        if (empty($searchFields)) {
             throw new InvalidArgumentException(
                 '$criterion->target',
                 "No searchable fields found for the given criterion target '{$criterion->target}'."
@@ -60,11 +60,12 @@ class FieldRelation extends Field
         $criterion->value = (array)$criterion->value;
 
         $queries = array();
-        foreach ($criterion->value as $value) {
-            $preparedValue = $this->escapeQuote($this->toString($value), true);
-
-            foreach ($fieldNames as $name) {
-                $queries[] = $name . ':"' . $preparedValue . '"';
+        foreach ($searchFields as $name => $fieldType) {
+            foreach ($criterion->value as $value) {
+                $preparedValues = (array)$this->mapSearchFieldvalue($fieldType, $value);
+                foreach ($preparedValues as $prepValue) {
+                    $queries[] = $name . ':"' . $this->escapeQuote($prepValue, true) . '"';
+                }
             }
         }
 
--- a/lib/Query/Content/CriterionVisitor/FullText.php
+++ b/lib/Query/Content/CriterionVisitor/FullText.php
@@ -44,9 +44,9 @@ class FullText extends CriterionVisitor
      *
      * @return array
      */
-    protected function getFieldNames(Criterion $criterion, $fieldDefinitionIdentifier)
+    protected function getSearchFields(Criterion $criterion, $fieldDefinitionIdentifier)
     {
-        return $this->fieldNameResolver->getFieldNames($criterion, $fieldDefinitionIdentifier);
+        return $this->fieldNameResolver->getFieldTypes($criterion, $fieldDefinitionIdentifier);
     }
 
     /**
@@ -76,9 +76,9 @@ class FullText extends CriterionVisitor
         );
 
         foreach ($criterion->boost as $field => $boost) {
-            $fieldNames = $this->getFieldNames($criterion, $field);
+            $searchFields = $this->getSearchFields($criterion, $field);
 
-            foreach ($fieldNames as $name) {
+            foreach ($searchFields as $name => $fieldType) {
                 $queries[] = $name . ':' . $criterion->value . '^' . $boost;
             }
         }
--- a/lib/Query/Content/CriterionVisitor/MapLocation.php
+++ b/lib/Query/Content/CriterionVisitor/MapLocation.php
@@ -56,7 +56,7 @@ abstract class MapLocation extends CriterionVisitor
     }
 
     /**
-     * Get field names.
+     * Get array of search fields.
      *
      * @param \eZ\Publish\API\Repository\Values\Content\Query\Criterion $criterion
      * @param string $fieldDefinitionIdentifier
@@ -65,13 +65,13 @@ abstract class MapLocation extends CriterionVisitor
      *
      * @return array
      */
-    protected function getFieldNames(
+    protected function getSearchFields(
         Criterion $criterion,
         $fieldDefinitionIdentifier,
         $fieldTypeIdentifier = null,
         $name = null
     ) {
-        return $this->fieldNameResolver->getFieldNames(
+        return $this->fieldNameResolver->getFieldTypes(
             $criterion,
             $fieldDefinitionIdentifier,
             $fieldTypeIdentifier,
--- a/lib/Query/Content/CriterionVisitor/MapLocation/MapLocationDistanceIn.php
+++ b/lib/Query/Content/CriterionVisitor/MapLocation/MapLocationDistanceIn.php
@@ -52,14 +52,14 @@ class MapLocationDistanceIn extends MapLocation
         $location = $criterion->valueData;
         $criterion->value = (array)$criterion->value;
 
-        $fieldNames = $this->getFieldNames(
+        $searchFields = $this->getSearchFields(
             $criterion,
             $criterion->target,
             $this->fieldTypeIdentifier,
             $this->fieldName
         );
 
-        if (empty($fieldNames)) {
+        if (empty($searchFields)) {
             throw new InvalidArgumentException(
                 '$criterion->target',
                 "No searchable fields found for the given criterion target '{$criterion->target}'."
@@ -68,7 +68,7 @@ class MapLocationDistanceIn extends MapLocation
 
         $queries = array();
         foreach ($criterion->value as $value) {
-            foreach ($fieldNames as $name) {
+            foreach ($searchFields as $name => $fieldType) {
                 $queries[] = sprintf('geodist(%s,%F,%F):%s', $name, $location->latitude, $location->longitude, $value);
             }
         }
--- a/lib/Query/Content/CriterionVisitor/MapLocation/MapLocationDistanceRange.php
+++ b/lib/Query/Content/CriterionVisitor/MapLocation/MapLocationDistanceRange.php
@@ -62,14 +62,14 @@ class MapLocationDistanceRange extends MapLocation
             $start = null;
         }
 
-        $fieldNames = $this->getFieldNames(
+        $searchFields = $this->getSearchFields(
             $criterion,
             $criterion->target,
             $this->fieldTypeIdentifier,
             $this->fieldName
         );
 
-        if (empty($fieldNames)) {
+        if (empty($searchFields)) {
             throw new InvalidArgumentException(
                 '$criterion->target',
                 "No searchable fields found for the given criterion target '{$criterion->target}'."
@@ -80,7 +80,7 @@ class MapLocationDistanceRange extends MapLocation
         $location = $criterion->valueData;
 
         $queries = array();
-        foreach ($fieldNames as $name) {
+        foreach ($searchFields as $name => $fieldType) {
             // @todo in future it should become possible to specify ranges directly on the filter (donut shape)
             $query = sprintf('{!geofilt sfield=%s pt=%F,%F d=%s}', $name, $location->latitude, $location->longitude, $end);
             if ($start !== null) {
--- a/lib/Resources/config/container/solr/criterion_visitors.yml
+++ b/lib/Resources/config/container/solr/criterion_visitors.yml
@@ -223,6 +223,7 @@ services:
         class: %ezpublish.search.solr.query.content.criterion_visitor.field_in.class%
         arguments:
             - @ezpublish.search.common.field_name_resolver
+            - @ezpublish.search.solr.field_value_mapper.aggregate
         tags:
             - {name: ezpublish.search.solr.query.content.criterion_visitor}
 
@@ -230,6 +231,7 @@ services:
         class: %ezpublish.search.solr.query.content.criterion_visitor.field_relation.class%
         arguments:
             - @ezpublish.search.common.field_name_resolver
+            - @ezpublish.search.solr.field_value_mapper.aggregate
         tags:
             - {name: ezpublish.search.solr.query.content.criterion_visitor}
 
@@ -237,6 +239,7 @@ services:
         class: %ezpublish.search.solr.query.content.criterion_visitor.field_range.class%
         arguments:
             - @ezpublish.search.common.field_name_resolver
+            - @ezpublish.search.solr.field_value_mapper.aggregate
         tags:
             - {name: ezpublish.search.solr.query.content.criterion_visitor}
 
@@ -384,6 +387,7 @@ services:
         class: %ezpublish.search.solr.query.location.criterion_visitor.field_in.class%
         arguments:
             - @ezpublish.search.common.field_name_resolver
+            - @ezpublish.search.solr.field_value_mapper.aggregate
         tags:
             - {name: ezpublish.search.solr.query.location.criterion_visitor}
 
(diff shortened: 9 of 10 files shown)


andrerom:
@pspanja Would like your input on this, in regards to what you had in mind here ;)

lib/Query/Content/CriterionVisitor/Field/FieldRange.php
```diff
@@ -71,7 +76,16 @@ public function visit(Criterion $criterion, CriterionVisitor $subVisitor = null)
         }
 
         $queries = array();
-        foreach ($fieldNames as $name) {
```
joaoinacio: ( `getFieldNames()` shouldn't be necessary any longer here, and the `empty()` check can be done on `$fieldTypes` instead )

bdunogier:
Really not a big fan of the type based check here. 

We do the same conversion work in `EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper\DateMapper::map()`. Could we somehow re-use that ? It has a check for the FieldType class. Not like we had that immediately available, but maybe we can work from that ?

pspanja:
This might be a bit bigger issue than it would seem :) Note that Field criterion handlers in Solr engine doesn't perform any mapping of the value given in the criterion. It seems that the motivation for reporting this was having queries that worked on Legacy engine breaking on Solr engine. Basically, in Legacy Storage date is stored as a timestamp/integer, while in Solr Search Engine it is indexed as a date field type, which can't be targeted with a integer value.

So we really need to implement proper input value mapping, more than the basic casting done in abstract [CriterionVisitor](https://github.com/ezsystems/ezplatform-solr-search-engine/blob/master/lib/Query/CriterionVisitor.php#L106). I think we can and should do that.

This approach has few problems, 1. it handles only range operators and 2. it assumes the type of targeted value for the field type. It would be better to have this in a dedicated service and dispatch mapping by the search [FieldType](https://github.com/ezsystems/ezpublish-kernel/tree/master/eZ/Publish/SPI/Search/FieldType), as suggested by @bdunogier. I think it should be possible to use `FieldValueMapper`, we should just find a way to expose the method for converting single value, so it works for multivalued fields. Also, as we'll need both field name and type, we'll probably need to adapt `FieldNameResolver` a bit, which is located in [`ezpublish-kernel`](https://github.com/ezsystems/ezpublish-kernel/tree/master/eZ/Publish/Core/Search/Common).

Single question I see is possibility to target fields of different type (if they have same field definition identifier), if going for this we should probably first decide how that should be handled.

joaoinacio:
So, if I understand correctly:
- modify FieldNameResolver to return an array with FieldType SearchField - check.
- use FieldValueMapper( aggregate?) to perform the SingleValued map from SPI Search FIeld - check.
- modify CriterionVisitors to use a map method (each multi-valued needs a slightly different way for calling that) - check.

My problem is: adding a map method to the base Field CriterionVisitor is simple, but is would require a dependency on the aggregate FieldValueMapper - would that be OK?

Also:
```
        $definition = $searchFieldType->getIndexDefinition();
        $searchField = new SearchField('field', $value, $definition['value']);
        return $this->aggregateFieldValueMapper->map($searchField);
```

comments?

joaoinacio:
Updated with the above concept, kernel PR for `FieldNameResolver` also updated.

A new service wasn't created as it seems this only needs to apply to the `Field` criterion visitors, so the mapping is done there for simplicity.

Also again for simplicity no change to FieldValueMappers, instead as noted above an SPI `Search\Field` is created from the indexable field definition and passed to the mapper.

AFAICT this approach seems to resolve the issue without the drawbacks of the original PR, even if there may be room for future improvement.

ping @pspanja @bdunogier @andrerom

lib/Query/Content/CriterionVisitor/Field/FieldIn.php
```diff
@@ -59,12 +59,12 @@ public function visit(Criterion $criterion, CriterionVisitor $subVisitor = null)
...
         $criterion->value = (array)$criterion->value;
-
         $queries = array();
-        foreach ($criterion->value as $value) {
-            $preparedValue = $this->escapeQuote($this->toString($value), true);
 
-            foreach ($fieldNames as $name) {
+        foreach ($searchFields as $name => $fieldType) {
+            foreach ($criterion->value as $value) {
+                $preparedValue = escapeQuote($this->mapSearchField($fieldType, $value), true);
```
pspanja: Missing `$this->` when calling `escapeQuote()`.

pspanja:
Quite neat :) Then, when targeting multiple fields of different type, we simply map the value for each one.
I think that's reasonable.

The tests didn't run, but this depends on https://github.com/ezsystems/ezpublish-kernel/pull/1551 right?
I think changing `ezpublish-kernel` dependency in `composer.json` to `dev-branch` would do the trick, but the PR there is from your fork, so that would not work.

lib/Query/Content/CriterionVisitor/Field.php
```diff
@@ -27,36 +30,52 @@
     protected $fieldNameResolver;
 
     /**
+     * Field value mapper.
```
bdunogier: I'd remove that, it just repeats the variables' name.

andrerom:
@joaoinacio take the kernel one so we can merge it to move forward on this, be aware you should target 6.1 branch with it in kernel for final review.

joaoinacio:
Replace remaining uses of `FIeldNameResolver::getFieldNames()` (MapUocation, FullText).

Note that no actual mapping is done (AFACT not necessary).
Otherwise, the mapping should probably be moved to a helper (from `Field` criterion visitor) instead.

joaoinacio:
ping @andrerom @pspanja

andrerom:
Looks okish to me, but ok, this means we need to 1. backport the kernel change to maintain 5.4 support, 2. tag that on 5.4.x and 6.1.x, and 3. bump requirements on kernel here for the tags.

lib/Query/Content/CriterionVisitor/Field.php
```diff
@@ -27,36 +30,50 @@
...
 
     /**
+     * @var \EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper
+     */
+    protected $fieldValueMapper;
+
+    /**
      * Create from content type handler and field registry.
      *
      * @param \eZ\Publish\Core\Search\Common\FieldNameResolver $fieldNameResolver
```
bdunogier: Missing phpdoc update

lib/Query/Content/CriterionVisitor/Field.php
```diff
@@ -27,36 +30,50 @@
...
             $criterion,
-            $fieldDefinitionIdentifier,
-            $fieldTypeIdentifier,
-            $name
+            $criterion->target
         );
     }
+
+    /**
+     * Map search field value to solr value using FieldValueMapper.
```
bdunogier: We can't reference solr in a generic part of the API.

We should I think assume that the `FieldValueMapper` we receive will change depending on the search engine being used.

lib/Query/Content/CriterionVisitor/FullText.php
```diff
@@ -76,9 +76,9 @@ public function visit(Criterion $criterion, CriterionVisitor $subVisitor = null)
         );
 
         foreach ($criterion->boost as $field => $boost) {
-            $fieldNames = $this->getFieldNames($criterion, $field);
+            $searchFields = $this->getSearchFields($criterion, $field);
 
-            foreach ($fieldNames as $name) {
+            foreach ($searchFields as $name => $fieldType) {
```
bdunogier: You could use `foreach (array_keys($searchFields) as $name)`

lib/Query/Content/CriterionVisitor/Field.php
bdunogier: Actually, I completely forgot that this was in the solr bundle itself. Please ignore the above.

lib/Query/Content/CriterionVisitor/FullText.php
joaoinacio: I used `foreach($arr as $key => $value)` in order to keep consistency between all uses in different criterion visitors, also more future-proof if the fieldtype happens to become needed for some reason IMO.

lib/Query/Content/CriterionVisitor/Field.php
```diff
@@ -27,36 +30,51 @@
...
     /**
+     * @var \EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper
+     */
+    protected $fieldValueMapper;
+
+    /**
      * Create from content type handler and field registry.
      *
      * @param \eZ\Publish\Core\Search\Common\FieldNameResolver $fieldNameResolver
+     * @param \EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper $fieldValueMapper
```
joaoinacio: done.

lib/Query/Content/CriterionVisitor/FullText.php
bdunogier: > more future-proof if the fieldtype happens to become needed for some reason IMO.

I don't really agree with this. If you change how the key's used, you can change the foreach. This has the drawback of leaving an unused variable. Anyway, not a huge deal.

bdunogier:
I'd like to get tests to run first. I guess they also fail on master, so I'll try to fix them there.

joaoinacio:
Hmm.. Tests always fail on composer install due to memory limit being reached, nothing I can do here AFAIK

bdunogier:
Opened https://github.com/ezsystems/ezplatform-solr-search-engine/pull/38, let's see if it passes. If it does, i'll merge, and you can rebase your PR on top of latest master.

bdunogier:
@joaoinacio memory issues fixed in master, can you rebase & push please ? There is a failure, but we'll see if we have _more_ with this change ;-)

joaoinacio:
The failure seems to be on kernel's FieldNameResolver, for some reason the contentTypeHandler is `null` at 
```
     protected function getSearchableFieldMap()
     {
         return $this->contentTypeHandler->getSearchableFieldMap();
     }
```
seems unrelated to the changes in ezsystems/ezpublish-kernel/pull/1551 , @bdunogier any hints?

bdunogier:
Looking into it.

bdunogier:
Got [3 jobs to pass out of 4](https://travis-ci.org/ezsystems/ezplatform-solr-search-engine/builds/114823747) on `master`. The remaining failure is still solr disappearing, despite having restarted it. It fails at `...............................................SS........FEFE 5734 / 8826 ( 64%)`, I'll restart it, and we can see if it it fails at the exact same spot again or not.

That aside, the failure seems normal to me. Unit tests pass on master, and fail on your branch on the [`FullTextTest`](https://github.com/ezsystems/ezplatform-solr-search-engine/blob/master/tests/lib/Search/Query/CriterionVisitor/FullTextTest.php), in `testVisitBoost()`. `FullText::visit()` is called, and calls `getSearchFields()`.

The  `FieldNameResolver` is [mocked](https://github.com/joaoinacio/ezplatform-solr-search-engine/blob/EZP-25309_invalid_date_string/tests/lib/Search/Query/CriterionVisitor/FullTextTest.php#L26), but only mocks the `getFieldName()` method. The newly added `getSearchFields()` is not mocked, tries to use the (unmocked) `contentHandler`, and fails.

I'd probably mock the method as well, and make it return what is expected from the method. Do you have enough to get it fixed ?

P.S. The [remaining job on master has passed](https://travis-ci.org/ezsystems/ezplatform-solr-search-engine/jobs/114823749), it seems.

joaoinacio:
Thanks, looking into it.

bdunogier:
Please use a distinct commit when handing reviews (`git commit --fixup=` is your friend).

Anyway, the mock looks good, let's see what travis thinks.

joaoinacio:
did now, didn't have the kernel error for some reason... dev-master a05d0e6 vs 6.0.0.2 on ci?

bdunogier:
> dev-master a05d0e6 vs 6.0.0.2 on ci?

I'm sorry, what ?

joaoinacio:
Job https://travis-ci.org/ezsystems/ezplatform-solr-search-engine/jobs/115106837 fails with a fatal error:
```PHP Fatal error:  Call to undefined method eZ\Publish\Core\Search\Common\FieldNameResolver::getFieldTypes() in /home/travis/build/ezsystems/ezplatform-solr-search-engine/lib/Query/Content/CriterionVisitor/Field.php on line 58```
This seems to be caused by the version of ezpublish-kernel used: `6.0.0.2` (I have `dev-master` locally)

**Comment by andrerom** [comment]
@joaoinacio you need to bump kernel requirements to 5.4.6 and 6.2 in composer.json, with that it will be easier to see which lib is causing it to pull in older kernel.

**Comment by andrerom** [comment]
@joaoinacio And remove this line: https://github.com/ezsystems/ezplatform-solr-search-engine/commit/f335f4613ea7d0ec2a889c9dc399f8ccbe26e37d

it is probably causing this.

**Comment by joaoinacio** [comment]
Updated to use `getDefaultMatchField()` instead of hardcoded `'value', which was causing problems in multivalued fields (f.e. Author does not have a 'value').

@andrerom: will do, which line in the commit?

**Comment by andrerom** [comment]
`"zendframework/zend-code": "~2.4.3"`

However this means we can't support PHP 5.4 anymore because of kernel dependency for testing, so this PR should bump version to v1.1, and I suggest the following composer.json changes:
```diff
diff --git a/.travis.yml b/.travis.yml
index 18b17b9..2931668 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
     include:
         - php: 5.6
           env: TEST_CONFIG="phpunit.xml"
-        - php: 5.4
+        - php: 7.0
           env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="4.10.4" CORES_SETUP="dedicated" SOLR_CONFS="lib/Resources/config/solr/schema.xml lib/Resources/config/solr/custom-fields-types.xml lib/Resources/config/solr/language-fieldtypes.xml" SOLR_CORES="core0 core1 core2 core3"
         - php: 5.5
           env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="4.10.4" CORES_SETUP="shared"    SOLR_CONFS="lib/Resources/config/solr/schema.xml lib/Resources/config/solr/custom-fields-types.xml lib/Resources/config/solr/language-fieldtypes.xml" SOLR_CORES="core0 core1 core2 core3"
diff --git a/composer.json b/composer.json
index 8394ccd..ef7f330 100644
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,7 @@
 {
     "name": "ezsystems/ezplatform-solr-search-engine",
     "description": "Solr search engine implementation for eZ Platform",
+    "homepage": "https://github.com/ezsystems/ezplatform-solr-search-engine",
     "license": "GPL-2.0",
     "authors": [
         {
@@ -9,12 +10,12 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "~5.4.5@dev | ~6.0@dev"
+      "php": "~5.5|~7.0",
+        "ezsystems/ezpublish-kernel": "~5.4.6@dev|~6.2@dev"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.7",
-        "matthiasnoback/symfony-dependency-injection-test": "0.*",
-        "zendframework/zend-code": "~2.4.3"
+        "matthiasnoback/symfony-dependency-injection-test": "0.*"
     },
     "autoload": {
         "psr-4": {
@@ -26,7 +27,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.0.x-dev"
+            "dev-master": "1.1.x-dev"
         }
     }
 }
```

bdunogier:
Great @andrerom, thank you.

bdunogier:
:+1:

andrerom:
_(we can handle php 7.0 and the dedicated tests failure as followup to this PR btw)_

andrerom:
@joaoinacio I'm having second thoughts on the php requirements bump, it means we potentially don1t solve the issue for 5.4 customers who reported it.

So I guess we will just have to skip testing on php 5.4 for the time being until we have a solution for this, but also change ""php": "~5.5|~7.0"," to ""php": "~5.4|~7.0",".

@bdunogier should this bump version number or not (1.0 to 1.1)?

bdunogier:
> @bdunogier should this bump version number or not (1.0 to 1.1)?

Yes, it should (new, backward compatible, features)

About php 5.4, let's summarize:
- if we ditch php 5.4 support, it won't be installable on eZ Publish Platform 5.4 with PHP 5.4 (that we do support)
- if we do not ditch it, we can't test php 5.4 because we don't have a publicly available version of kernel we can test with

Did I get this right ?

If yes, I'd say no automated testing (at the moment) on PHP 5.4, but we do manual testing for eZ 5.4.

joaoinacio:
Fixed remaining error (array to string conversion) caused by multi-valued fields such as author etc (fieldValueMapper may return an array so use first element), tests should pass now.

Regarding PHP  temporarily disabling test on 5.4 seems reasonable AFAICT.
Should I go ahead with last change and bump version ?

andrerom:
> Should I go ahead with last change and bump version ?

yes, I have updated patch suggestion above with what we ended up with.

> Yes, it should (new, backward compatible, features)

@bdunogier this is bug, but yes it was solved by requiring a new feature in kernel, so technically this could still be 1.0.x here, but I don't mind bumping it because of requirement change.

bdunogier:
+1 for bump because of requirements changes, yes.

andrerom:
@joaoinacio you forgot the `"zendframework/zend-code": "~2.4.3"` part, otherwise looks good, hopefully it will pass with this :)

joaoinacio:
Travis seems to be having issues with solr going down (?), not sure what - if anything - can be done regarding that?

andrerom:
> Travis seems to be having issues with solr going down (?), not sure what - if anything - can be done regarding that?

nothing we can do with that other then maybe switching to container env

andrerom:
But maybe you should run the integration and unit test locally first, and using composer update to pull in right dependencies , cause seems you still have some array to string issues.

andrerom:
You can rebase on top of master which is now green, now and rather focus on stabilizing this _(ref #41)_ ;)

andrerom:
@joaoinacio https://github.com/ezsystems/ezplatform-solr-search-engine#testing-locally ;)

wizhippo:
Any chance this will be updated soon? I need date support.

andrerom:
@wizhippo I think @joaoinacio is a bit blocked on the remaining failures (see  #41), you may have a look if you want.

wizhippo:
FYI rebased `joaoinacio:EZP-25309_invalid_date_string` on master and ran tests locally. 

```
OK, but incomplete, skipped, or risky tests!
Tests: 8806, Assertions: 25161, Skipped: 48, Incomplete: 2.
```

andrerom:
did you run integration tests? *(those are the once failing https://travis-ci.org/ezsystems/ezplatform-solr-search-engine/builds/115527164)*

andrerom:
Locally here on single core setup on php 7.0.4, with dev-master kernel, I'm getting:
```
Tests: 8806, Assertions: 24719, Errors: 6, Failures: 176, Skipped: 48, Incomplete: 2.
```

[On travis with dedicated codes, php 5.4, and with kernel v6.0.1.2:](https://travis-ci.org/ezsystems/ezplatform-solr-search-engine/jobs/115527166)
```
Tests: 8796, Assertions: 24665, Errors: 6, Failures: 176, Skipped: 48, Incomplete: 2.
```

[On travis with shared codes, php 5.5, and with kernel v6.2.0:](https://travis-ci.org/ezsystems/ezplatform-solr-search-engine/jobs/115527167)
```
Tests: 8806, Assertions: 24719, Errors: 6, Failures: 176, Skipped: 48, Incomplete: 2.
```

[On travis with single code, php 5.6, and with kernel v6.2.0:](https://travis-ci.org/ezsystems/ezplatform-solr-search-engine/jobs/115527168)
```
Tests: 8806, Assertions: 24719, Errors: 6, Failures: 176, Skipped: 48, Incomplete: 2.
```

On all it is "Array to string conversion" causing it to fail.
Did you run it with something else perhaps?

wizhippo:
Odd.  Locally I have the same packages as yourself https://gist.github.com/wizhippo/51076d6fe092de85738c/revisions.

I followed step 1-5 from https://github.com/ezsystems/ezplatform-solr-search-engine#testing-locally. 

Here is the output when I run the tests https://gist.github.com/wizhippo/ee44284c259dee25d6b7

 Do I have to do anything else as in my output I see `This test depends on "PHPUnit_Framework_SkippedTestCase::testCreateTestContent" to pass.`?

Here is my rebased repo https://github.com/wizhippo/ezplatform-solr-search-engine/commit/d18aabd3af4c96eb755e6faf27b06de8a7d83a4d

lib/Query/Content/CriterionVisitor/Field.php
```diff
@@ -27,36 +30,55 @@
...
+     *
+     * @param eZ\Publish\SPI\FieldType\Indexable $searchFieldType
+     * @param mixed $value
+     *
+     * @return mixed
+     */
+    protected function mapSearchFieldValue(Indexable $searchFieldType, $value)
+    {
+        $definition = $searchFieldType->getIndexDefinition();
+        $matchField = $searchFieldType->getDefaultMatchField();
```
wizhippo: This seem to be the reason for the failures.  `getDefaultMatchField()` is returning `selected_option_index` even when `selected_option_value` is expected.  How can we distinguish which `indexDefinition` should be used?

lib/Query/Content/CriterionVisitor/Field/FieldRelation.php
```diff
@@ -60,10 +60,10 @@ public function visit(Criterion $criterion, CriterionVisitor $subVisitor = null)
         $criterion->value = (array)$criterion->value;
 
         $queries = array();
-        foreach ($criterion->value as $value) {
-            $preparedValue = $this->escapeQuote($this->toString($value), true);
+        foreach ($searchFields as $name => $fieldType) {
+            foreach ($criterion->value as $value) {
+                $preparedValue = $this->escapeQuote($this->mapSearchFieldvalue($fieldType, $criterion->value), true);
```
wizhippo: should this not be just `$value` instead of `$criterion->value`

lib/Query/Content/CriterionVisitor/Field.php
wizhippo: Adding this test specific to Selection fields passes local tests, but we add again a hard coded dependency.
How can we address this otherwise?  Is the `defaultMatchField` incorrect for this type?

```
        if ($searchFieldType instanceof \eZ\Publish\Core\FieldType\Selection\SearchField) {
            $matchField = 'selected_option_value';
        } else {
            $matchField = $searchFieldType->getDefaultMatchField();
        }
```

joaoinacio: I have the same question, not completely certain if the match field is incorrect here...

pspanja: Default match field is used as default field with Field criterion, while other fields indexed by the field type can be targeted through `CustomFieldInterface` implemented by it and with `CustomField` criterion. Since this is `Field` criterion, I would guess `CustomFieldInterface` is not handled here. Sorry I don't have time ATM to look more into it.

bdunogier:
Aaah, all green, good job guys.

@joaoinacio please make sure you don't amend / rebase when attempting to fix these. New commits allow us to see upon which change it started passing. It makes reviews easier (we can squash on multiple occurences when a pull-request takes a long time to get accepted).

lib/Query/Content/CriterionVisitor/Field.php
```diff
@@ -27,36 +30,60 @@
...
+     * @param eZ\Publish\SPI\FieldType\Indexable $searchFieldType
+     * @param mixed $value
+     *
+     * @return mixed
+     */
+    protected function mapSearchFieldValue(Indexable $searchFieldType, $value)
+    {
+        $definition = $searchFieldType->getIndexDefinition();
+        if ($searchFieldType instanceof \eZ\Publish\Core\FieldType\Selection\SearchField) {
+            $matchField = 'selected_option_value';
```
pspanja: This should not be hardcoded. Key in the array returned by `getSearchFields()` is the field name that should be used, it should be passed here as match field. Then you can lose this check and just use that.

But I'm not sure why `FieldNameResolver::getFieldTypes()` returns `Indexable` implementations as values, it seems like search field type should be enough. BTW you call it that here, so that was probably the idea. Then you could also have `$fieldType` below passed here, and there should be no need to find it through `getIndexDefinition()`.

wizhippo: The key returned by getSearchFields is already decorated like `test-ezselection_data_selected_option_value_ms` and on the getIndexDefintion the key is `selected_option_value` so it can't be used as is.

pspanja: Right, missed that, but with search field type instead of `Indexable` you wouldn't even need it here.

bdunogier:
Let's clarify what @pspanja and @wizhippo discussed before we merge.

lib/Query/Content/CriterionVisitor/Field.php
wizhippo: Sorry, it must be so obvious I'm just not seeing it.  How do we go from `getSearchFields` that is an array of `Indexable` to `Search\FieldType`?

pspanja: `FieldNameResolver` was changed in https://github.com/ezsystems/ezpublish-kernel/pull/1551 for the purpose of this PR, so the answer is there.

Additional problem is how custom fields are handled, in this case you can't know the search field type for it. Current implementation will just use the type of the default match field, I guess that is why special handling for Selection was needed, which basically just uses the string type field for mapping, over the default field which is of integer type. TBH I'm not completely sure why the same case doesn't fail elsewhere, probably because default field is of string type, so values are being mapped to string.

wizhippo: `getSearchFields` still returns `Indexable`'s after that PR, so you can't just use that searchtype to create the SearcField. Maybe there is another idea there I just don't see.

An alternative that is a little better then the hard coded value but not much would be to iterate through the definition keys and see if they exist in the name of the fieldtype as done here https://github.com/wizhippo/ezplatform-solr-search-engine/commit/0a19a5c1d655d061a972051d3d3a368a08ce32fd.  This way at least it correctly picks the fieldtype for `Selection\SearchField` for both value and index. As it stands now it always uses value.  Not sure, but I think a test should have failed if the wrong index `test-ezselection_data_selected_option_value_ms` vs `test-ezselection_data_selected_option_index_mi` is used.

pspanja: > getSearchFields still returns Indexable's after that PR, so you can't just use that searchtype to create the SearcField. Maybe there is another idea there I just don't see.

The idea is that `FieldNameResolver` should be changed.

> An alternative that is a little better then the hard coded value but not much would be to iterate through the definition keys and see if they exist in the name of the fieldtype as done here wizhippo@0a19a5c.

That might work for the moment, but it's error prone, as you can't know how the field name string looks like and if the correct part of it will be matched.

> Not sure, but I think a test should have failed if the wrong index test-ezselection_data_selected_option_value_ms vs test-ezselection_data_selected_option_index_mi is used.

It will work because the value always ends up being converted to string as part of the Solr query string.
BTW just noticed that `CriterionVisitor::toString()` is not called on the mapped value either.

wizhippo: I've added the use of CriterionVisitor::toString() to mine.

Thanks for helping me out.  I'll let you guys figure it out then.  It's a little more involved then I have time for at the moment.

wizhippo: > Additional problem is how custom fields are handled, in this case you can't know the search field type for it

This is the major issue I've run into. Changing `FieldNameResolver::getFieldTypes` to return SearchFieldType's is simple enough by getting it from the definition like:

```
            if ($name === null) {
                $name = $fieldType->getDefaultMatchField();
            }
            
            $definition = $fieldType->getIndexDefinition();

            $fieldTypeNameMap[$fieldName] = $definition[$name];
```

This works except for `test-ezselection_data_selected_option_value_ms` as it used the `getDefaultMatchField`and uses index instead as was pointed out to be a problem in a previous remark.  Some how the correct name to be used for the definition key can be retrieved or passed in to `getFieldTypes` from the visitor's.

I'm really done digging in for now:)  Looking forward to see how it's done.

pspanja: > This is the major issue I've run into.

Do you mean `CustomField` criterion, or using `CustomFieldInterface` on the `Field` criterion?
I was thinking about the latter, how useful in real world it really is, because while it does provide override by Content Type, it also prevents some other quite useful stuff, like sorting on a field of the same type and with the same name across Content Types. TBH that seems like a more important case to me.

> Some how the correct name to be used for the definition key can be retrieved or passed in to getFieldTypes from the visitor's.

If you don't want the default field, then you have to know exactly what field to use. This is already done in `MapLocation` criterion visitor, where the field name is injected through the constructor, because the default field used with `Field` criterion is not geo location, but an address string. I was thinking about exposing available field names through constants in `Indexable` definition, then you would be able to use that to get the field name in Solr index. The method to do it should be implemented in `FieldNameResolver`, but most of the logic is already there in the `getIndexField()` method.

I think we should eventually have a helper for that.

wizhippo: `CustomFieldInterface` I believe.  For example in the SelectionIntegationTests where there is a `Field` with `customFields` of `test-ezselection` which has a `data` property of `test-ezselection_data_selected_option_value_ms`.  How is the `FieldIn` visitor supposed to to know the name to use is `selected_option_value` vs the default match `selected_option_index` in this case?  The example you sited `MapLocation` has a fieldName property, Field does not.

joaoinacio:
@pspanja: given the discussion, could you please clarify/elaborate on the changes you feel would be necessary to move this forward?

> If you don't want the default field, then you have to know exactly what field to use.

Currently I share the same concerns already mentioned by @wizhippo

>  How is the FieldIn visitor supposed to to know the name to use is selected_option_value vs the default match selected_option_index in this case?

bdunogier:
Is there something blocking this from being merged ?

andrerom:
This: https://github.com/ezsystems/ezplatform-solr-search-engine/pull/35#discussion_r56888072

bdunogier:
Joao, could you please try to reformulate the current question, linked by André above, as a todo item on the PR ? It would help keeping track of the status.

wizhippo:
Anything I can do to help move this along?  Searching by date is a pretty important feature.  Unless I can find a way to resolve the correct index field name discussed above, I'm not sure what I can do to remove the hard coded field name, but If anyone can help point me int the right direction I'm happy to help.

pspanja:
Hi @wizhippo, interesting that you bring this up now, I was just working a bit on this together with @galileo and the plan is to have it fixed tomorrow :)

wizhippo:
@pspanja Great news. I'll be looking forward to see how you guys did it.

andrerom:
Closed for #45

PR #36, opened 2016-01-08
Indexer plugins feature

silverjgo:
I implemented a possibility to add plug-in services, which hook into the indexing process. These plug-ins are called per content object and language instead for every field. Also, they can add new Solr fields for content objects instead of overriding the behaviour of existing field types (via alternative Indexable service or FieldValueMapper). Thus, it provides a broader range of options for customization of the index.


lib/DocumentMapperPluginInterface.php
```diff
@@ -0,0 +1,46 @@
...
+     *
+     * @param Type $type
+     * @return bool
+     */
+    public function canExtend(Type $type);
+
+    /**
+     * This is called for every indexed content object and language.
+     *
+     * The generated Field instances SHOULD get a prefix (e.g. ext_) in their
```
andrerom: "The generated Field instances SHOULD get a prefix indicating it is extension & identified by provider (e.g. ext_ez_) in their"

pspanja:
Hi @silverjgo, this implementation is a bit limiting, for example you assume that one will only want to index additional fields per ContentType. What about indexing additional fields for every Content, or for content in a specific Section?
We should also have a way to index additional fields to the Location documents.
From document indexing plugin I would expect:

- Ability to target Content or Location document (then maybe different interfaces for each).
- Dispatching on any property of the Content or Location (`canExtend()` or similar would then accept `Content` or `Location` object).
- Checking for field name conflicts.
  Fields generated in `NativeDocumentMapper` should have precedence over fields added by the plugin. An exception should be thrown in case of conflict. Maybe it would also make sense to prefix fields added by plugins so they can be easily distinguished and conflicts are avoided.

lib/DocumentMapper/NativeDocumentMapper.php
```diff
@@ -364,6 +381,16 @@ public function mapContentBlock(Content $content)
             $isMainTranslation = ($content->versionInfo->contentInfo->mainLanguageCode === $languageCode);
             $alwaysAvailable = ($isMainTranslation && $content->versionInfo->contentInfo->alwaysAvailable);
 
+            $extensionFields = array();
+            foreach ($this->mapperPlugins as $plugin) {
+                if ($plugin->canExtend($contentType)) {
+                    $extensionFields = array_merge(
```
andrerom: Can use array union, and end up with something ala:
```php
$extensionFields += $plugin->createExtensionFields($content, $contentType, $languageCode);
```

andrerom:
> this implementation is a bit limiting, for example you assume that one will only want to index additional fields per ContentType. What about indexing additional fields for every Content, or for content in a specific Section?

indeed
What about having a completely generic visitors approach ala Criterion handlers and so on? And a tag + interface for content and one for location, something like:
```php
interface ContentDocumentPlugin
{
    public function canHandleContent(Content $content);
    public function getContentFields(Content $content, $languageCode);
}

interface LocationDocumentPlugin
{
    public function canHandleLocation(Location $location);
    public function getLocationFields(Location $location, Content $content);
}
```

pspanja:
That looks good, I'm just not sure how to handle language parameter. I mean it would be good to have that, but I'm thinking about the case when getting additional data is not cheap + we want it for every Content document and not per specific translation. Then the interface would force doing the same for every translation, or maybe implementing some kind of caching, which seems out of order.

Maybe we can lose `$languageCode` in `getContentFields()` and introduce additional method, `getFieldsForTranslation(Content $content, $languageCode)` or something like that.
What do you think?

andrerom:
Sure, so then interface would need to be something like `ContentTranslationDocumentPlugin` then, right?

andrerom:
*Side: It is a bit confusing that we use the term "field" for several meanings*

pspanja:
Hi @silverjgo, since nothing happened here for some time, I opened alternative proposal at https://github.com/ezsystems/ezplatform-solr-search-engine/pull/42, hope you won't mind :)

PR #38, opened 2016-03-09
[travis] fixed php memory limit issues

bdunogier:
Disabled xdebug and memory limit when running composer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,9 @@ branches:
 
 before_script:
     - composer selfupdate
+    - phpenv config-rm xdebug.ini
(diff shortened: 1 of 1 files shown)


andrerom:
https://github.com/ezsystems/LegacyBridge/commit/d2c5534fad165e5095a8b16f081ed2bd36bf5c4b

bdunogier:
Thank you @andrerom, I'll look into it if what I have just pushed fails as well.

bdunogier:
Okay, memory issues are gone. We have failures now :-)

bdunogier:
+1 for merging it, even if it fails ? It's still an improvement over the current {{master}}.

andrerom:
+1

PR #40, opened 2016-03-12
v1.1 Bump and testing improvements 

andrerom:
To unblock #35:
- Bump requirements and version numbers
  - tag kernel v6.0.1 and v6.1.1 with #35 enhancement, 6.0.x for testing on php 5.4 
- Add testing on PHP 7.0
- Switch to Travis Trusty environment for:
 - Stable runs _(no more lost connection to solr issues)_
 - More power _for faster integration runs_
- Improve how composer packages are installed _(prefer stable)_

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+## Run on trusty environment as solr dies all the time on containers after travis move to gce
+sudo: required
+dist: trusty
+
 language: php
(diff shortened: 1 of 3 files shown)


bdunogier:
Hmmm, did we need to bump to 1.1 for this ?

PR #42, opened 2016-04-25
Implement indexing plugins

pspanja:
This PR implements https://jira.ez.no/browse/EZP-24805

This implements document field mappers, aka indexing plugins feature.
Four types of base plugins are implemented:

1. `Content::mapFields(Content $content)`
2. `ContentTranslation::mapFields(Content $content, $languageCode)`
3. `Location::mapFields(Location $location)`
4. `LocationTranslation::mapFields(Location $location, $languageCode)`

For each of these an `Aggregate` plugin is implemented and used (6 places total):

1. `Content::mapFields(Content $content)`
  1. Block - indexes fields on the block document level -- meaning to each Content and Location document in the block
  2. Content - indexes fields to each Content document
2. `ContentTranslation::mapFields(Content $content, $languageCode)`
  1. BlockTranslation - indexes fields on the block level per translation
  2. ContentTranslation - indexes fields to Content document per translation
3. `Location::mapFields(Location $location)`
  1. Location - indexes field to each Location document
4. `LocationTranslation::mapFields(Location $location, $languageCode)`
  1. LocationTranslation - indexes field to Location document per translation

The same, but starting from extension points (service tags):

1. Block: `ezpublish.search.solr.document_field_mapper.block`
  - Accepts `Content::mapFields(Content $content)`
2. Block translation: `ezpublish.search.solr.document_field_mapper.block_translation`
  - Accepts `ContentTranslation::mapFields(Content $content, $languageCode)`
3. Content: `ezpublish.search.solr.document_field_mapper.content`
  - Accepts `Content::mapFields(Content $content)`
4. Content translation: `ezpublish.search.solr.document_field_mapper.content_translation`
  - Accepts `ContentTranslation::mapFields(Content $content, $languageCode)`
5. Location: `ezpublish.search.solr.document_field_mapper.location`
  - Accepts `Location::mapFields(Location $location)`
6. Location translation: `ezpublish.search.solr.document_field_mapper.location_translation`
  - Accepts `Content::mapFields(Location $location, $languageCode)`

Concrete plugins are registered to `Aggregate` plugin in a container compiler pass, using service tags.
At the moment no concrete plugins are implemented, followup to this would be refactoring `DocumentMapper` to use only plugins to map fields.

### TODOs
- [x] Get rid of argument unpacking (`...` operator) as it is not supported on PHP <5.6

### DIFF
--- a/lib/Container/Compiler/DocumentFieldMapperPass/ContentTranslation.php
+++ b/lib/Container/Compiler/DocumentFieldMapperPass/ContentTranslation.php
@@ -1,0 +1,23 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
+ */
+namespace EzSystems\EzPlatformSolrSearchEngine\Container\Compiler\DocumentFieldMapperPass;
+
+use EzSystems\EzPlatformSolrSearchEngine\Container\Compiler\BaseDocumentFieldMapperPass;
+
+/**
+ * Compiler pass for aggregate document field mapper for the Content document
+ * in a specific translation.
+ */
+class ContentTranslation extends BaseDocumentFieldMapperPass
+{
+    const AGGREGATE_MAPPER_SERVICE_ID = 'ezpublish.search.solr.document_mapper.plugin.content_translation';
+    const AGGREGATE_MAPPER_SERVICE_TAG = self::AGGREGATE_MAPPER_SERVICE_ID;
+}
--- a/lib/Resources/config/container/solr.yml
+++ b/lib/Resources/config/container/solr.yml
@@ -47,6 +47,12 @@ services:
     ezpublish.search.solr.document_mapper.native:
         class: %ezpublish.search.solr.document_mapper.native.class%
         arguments:
+            - @ezpublish.search.solr.document_field_mapper.block
+            - @ezpublish.search.solr.document_field_mapper.block_translation
+            - @ezpublish.search.solr.document_field_mapper.content
+            - @ezpublish.search.solr.document_field_mapper.content_translation
+            - @ezpublish.search.solr.document_field_mapper.location
+            - @ezpublish.search.solr.document_field_mapper.location_translation
             - @ezpublish.search.common.field_registry
             - @ezpublish.spi.persistence.content_handler
             - @ezpublish.spi.persistence.location_handler
--- a/lib/Resources/config/container/solr/services.yml
+++ b/lib/Resources/config/container/solr/services.yml
@@ -4,6 +4,12 @@ parameters:
     ezpublish.search.solr.query.common.sort_clause_visitor.aggregate.class: EzSystems\EzPlatformSolrSearchEngine\Query\Common\SortClauseVisitor\Aggregate
     ezpublish.search.solr.query.common.facet_builder_visitor.aggregate.class: EzSystems\EzPlatformSolrSearchEngine\Query\Common\FacetBuilderVisitor\Aggregate
     ezpublish.search.solr.field_value_mapper.aggregate.class: EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper\Aggregate
+    ezpublish.search.solr.document_field_mapper.block.class: EzSystems\EzPlatformSolrSearchEngine\DocumentMapper\FieldMapper\Content\Aggregate
+    ezpublish.search.solr.document_field_mapper.block_translation.class: EzSystems\EzPlatformSolrSearchEngine\DocumentMapper\FieldMapper\ContentTranslation\Aggregate
+    ezpublish.search.solr.document_field_mapper.content.class: EzSystems\EzPlatformSolrSearchEngine\DocumentMapper\FieldMapper\Content\Aggregate
+    ezpublish.search.solr.document_field_mapper.content_translation.class: EzSystems\EzPlatformSolrSearchEngine\DocumentMapper\FieldMapper\ContentTranslation\Aggregate
+    ezpublish.search.solr.document_field_mapper.location.class: EzSystems\EzPlatformSolrSearchEngine\DocumentMapper\FieldMapper\Location\Aggregate
+    ezpublish.search.solr.document_field_mapper.location_translation.class: EzSystems\EzPlatformSolrSearchEngine\DocumentMapper\FieldMapper\LocationTranslation\Aggregate
 
 services:
     ezpublish.search.solr.gateway.client.http.stream:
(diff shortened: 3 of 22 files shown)


pspanja:
Ping @andrerom @bdunogier

andrerom:
Cool.

does this replace #36?
For Repo/Solr stuff you can also start to ping @galileo btw :)

lib/Resources/config/container/solr.yml
```diff
@@ -47,6 +47,12 @@ services:
     ezpublish.search.solr.document_mapper.native:
         class: %ezpublish.search.solr.document_mapper.native.class%
         arguments:
+            - @ezpublish.search.solr.document_field_mapper.block
```
bdunogier: That's a LOT of dependencies for one service.

lib/Resources/config/container/solr/services.yml
```diff
@@ -43,3 +49,33 @@ services:
     # are registered to this one using compilation pass
     ezpublish.search.solr.query.location.facet_builder_visitor.aggregate:
         class: %ezpublish.search.solr.query.common.facet_builder_visitor.aggregate.class%
+
+    # Note: services tagged with 'ezpublish.search.solr.document_field_mapper.block'
```
bdunogier: I don't understand this comment. Where's that tag ? The comment is about a tag, but the service isn't tagged with it.

lib/Container/Compiler/DocumentFieldMapperPass/ContentTranslation.php
```diff
@@ -0,0 +1,23 @@
...
+
+use EzSystems\EzPlatformSolrSearchEngine\Container\Compiler\BaseDocumentFieldMapperPass;
+
+/**
+ * Compiler pass for aggregate document field mapper for the Content document
+ * in a specific translation.
+ */
+class ContentTranslation extends BaseDocumentFieldMapperPass
+{
+    const AGGREGATE_MAPPER_SERVICE_ID = 'wunderman_puls4.search.solr.document_mapper.plugin.content_translation';
```
bdunogier: I don't understand where this `wunderman_puls4` service name is coming from.

pspanja:
@andrerom 
> does this replace #36?

Yup, I left a comment there.

> For Repo/Solr stuff you can also start to ping @galileo btw :) 

Ok, good to know :)

lib/Container/Compiler/DocumentFieldMapperPass/ContentTranslation.php
pspanja: I'll need to remove that ;)

lib/Resources/config/container/solr/services.yml
pspanja: You should read the description ;)

lib/Resources/config/container/solr.yml
pspanja: Indeed, only field mappers will remain once `DocumentMapper` is refactored to use them. It's in the PR description.

lib/Container/Compiler/DocumentFieldMapperPass/ContentTranslation.php
pspanja: Amended and force pushed.

lib/Resources/config/container/solr/services.yml
bdunogier: If it requires that you read the pull-request's description to make sense out of the comment, then it is a poor comment.

pspanja: I honestly don't know what to say.

bdunogier:
Please @pspanja do not rebase on the same commit on every change, it makes reviews & notifications very tedious.

Use `commit --fixup`, and you'll `rebase --autosquash` when we've reached a consensus.

galileo:
Yeah, I will first try to do my best to align my knowledge about solr and this implementation itself.

lib/Resources/config/container/solr/services.yml
bdunogier: Okay, I got it a bit better.

Maybe what troubled me is that we don't usually add such comments, especially not in the services file. Also, the N comments in sequence are very similar, and it's not immediately obvious what they refer to. Maybe shorten them ?

`# block mappers are tagged as XXX` ?

andrerom:
> Use commit --fixup, and you'll rebase --autosquash when we've reached a consensus.

Could, but with squash support on github now it's not a must anymore, lets PR remain with PR history, while changes merged into master being clean.

bdunogier:
> Could, but with squash support on github now, so it's not a must anymore. It lets PR remain with PR history, while changes merged into master being clean.

The point was to stop amending to the PR's commit in order to avoid invalid links in notifications.

andrerom:
> The point was to stop amending to the PR's commit in order to avoid invalid links in notifications.

I know, but my point is that we actually don'

lib/Resources/config/container/solr/services.yml
pspanja: Comments like those are actually all over the place, I always used that format.

pspanja:
This should be clean now, aside for CS checker not recognizing `...` operator (5.6+).

andrerom:
> aside for CS checker not recognizing splat/scatter operator (..., 5.6+).

Variadic-functions/argument-unpacking. Be aware we are on PHP 5.4.4+ here, only possible bump for next major is PHP 5.5.9+ as defined by Symfony 3.0. So best if you don't use that yet ;)

pspanja:
> Be aware we are on PHP 5.4.4+ here

Ok, todo added.

pspanja:
Removed argument expansion in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/42/commits/042b297c0dc9a5a747b1d1400e28f4becbdf95ea.

pspanja:
Registered compiler passes with container (bundle and test setup) in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/42/commits/33a6fe6500dd31786faf00c35f8954c6faf1d09b.
This should be complete now.

pspanja:
So, good to go? :)

andrerom:
I'm pretty much good with it, but I'd like to see us do this with `followup to this would be refactoring DocumentMapper to use only plugins to map fields.` in one go, not in PR here, but as an PR for this one, hence making it possible to see what we end up with.

@bdunogier what do you think?

pspanja:
Sounds good to me, I already did some work on that to validate the whole idea. Also have several other follow-ups on the whole thing:

- syncing some Content and Location mappers (and criterion visitors) and replacing them with single Block mapper
- using all that work to move criterion visitors to Common namespace and thus remove some duplication we ended up with for historical reasons

pspanja:
Here we go: https://github.com/pspanja/ezplatform-solr-search-engine/pull/1

andrerom:
Thanks, should probably discuss this in person or in a call to better understand it, but:
- if we are to index location drafts and trash locations in the future, then we need to think a bit about that now to prepare thse extension points to always get the right locations.
- Will need to dig into this a bit and what you use this for, but if it was possible to center content fields *(non translated)* around ContentInfo it would make the separation more clear.
- Didn't fully get what `LocationTranslation::mapField` are for
- Neither the block mappers, we might need `doc/design/solr_document.rst` and a `doc/design/field_mappers.rst` files to explain how it works ;)

*Note: Only glanced on this so far, it's big so haven't gotten an overview yet.*

pspanja:
> if we are to index location drafts and trash locations in the future, then we need to think a bit about that now to prepare thse extension points to always get the right locations.

I'm not completely sure how that would look like, but seems like it should already be covered by the interface (fetching trashed/draft Locations by the Content). In that case we only miss the slots.

> Will need to dig into this a bit and what you use this for, but if it was possible to center content fields (non translated) around ContentInfo it would make the separation more clear.

I see what you mean, but `ContentInfo` does not contain all necessary data, for example creator user ID is held in `VersionInfo`. Besides, Content is the starting point.

> Didn't fully get what LocationTranslation::mapField are for

We don't have a use for it here, it is there for users to index stuff per translation to Locations only, if the use-case demands it. Maybe it would make sense to pass Content there as well.

> Neither the block mappers, we might need doc/design/solr_document.rst and a doc/design/field_mappers.rst files to explain how it works ;)

Block is basically nested documents unit, so Content + its Locations nested, per translation. Block needs to be indexed as a whole. By having a mapper on that level we can provide fields that go to both Content and Location documents in the block only once, both per translation (block translation point) and regardless of translation, meaning all Content's blocks (block point).

Anyways, just ping me on Skype or something if you want to discuss this.

pspanja:
Updated description to make that more clear, hopefully.

pspanja:
So what do we do with this :)
It started getting into my way, if you don't mind I'm going close it.

andrerom:
Lots of holiday last week.

>> Didn't fully get what LocationTranslation::mapField are for
> We don't have a use for it here, it is there for users to index stuff per translation to Locations only, if the use-case demands it. Maybe it would make sense to pass Content there as well.

Yes either skip it or pass Content, given Location isn't translated it does no make sense as-is.

pspanja:
Closing in favor of https://github.com/ezsystems/ezplatform-solr-search-engine/pull/67

PR #43, opened 2016-05-19
EZP-25788: Solr not indexing on content publish

janit:
Added comment about commit and softcommit configurations in Solr.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ For Contributing to this Bundle, you should make sure to run both unit and integ
     cp -R lib/Resources/config/solr/* solr-4.10.4/example/solr/collection1/conf
     ```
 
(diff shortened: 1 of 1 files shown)


andrerom:
@janit If you agree I think we can do better, could you add something like the following as well:
```
Example of working `solrconfig.xml` settings that you can use to tune for your needs *(changes here: enables `openSearcher` and changes AutoSoftCommit from `-1` disabled to `50`ms)*:

         <autoCommit> 
           <maxTime>${solr.autoCommit.maxTime:15000}</maxTime> 
           <openSearcher>true</openSearcher> 
         </autoCommit>

         <autoSoftCommit> 
           <maxTime>${solr.autoSoftCommit.maxTime:50}</maxTime> 
         </autoSoftCommit>
```

andrerom:
ping @pspanja

andrerom:
@janit discussed with @pspanja, and seems we don't need openSearcher, if softCommit is done, so updated comment above.

andrerom:
+1

pspanja:
+1

PR #45, opened 2016-05-20
EZP-25309: Fix datetime value mapping

galileo:
Issue: https://jira.ez.no/browse/EZP-25309
~~Depends on https://github.com/ezsystems/ezpublish-kernel/pull/1655~~

### DIFF
--- a/lib/Query/Content/CriterionVisitor/Field.php
+++ b/lib/Query/Content/CriterionVisitor/Field.php
@@ -15,7 +15,7 @@ use EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper;
 use eZ\Publish\Core\Search\Common\FieldNameResolver;
 use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
 use eZ\Publish\SPI\Search\Field as SearchField;
-use eZ\Publish\SPI\FieldType\Indexable;
+use eZ\Publish\SPI\Search\FieldType;
 
 /**
  * Visits the Field criterion.
@@ -64,26 +64,20 @@ abstract class Field extends CriterionVisitor
     /**
      * Map search field value to solr value using FieldValueMapper.
      *
-     * @param eZ\Publish\SPI\FieldType\Indexable $searchFieldType
      * @param mixed $value
      *
+     * @param \eZ\Publish\SPI\Search\FieldType $searchFieldType
      * @return mixed
      */
-    protected function mapSearchFieldValue(Indexable $searchFieldType, $value)
+    protected function mapSearchFieldValue($value, FieldType $searchFieldType = null)
     {
-        $definition = $searchFieldType->getIndexDefinition();
-        if ($searchFieldType instanceof \eZ\Publish\Core\FieldType\Selection\SearchField) {
-            $matchField = 'selected_option_value';
-        } else {
-            $matchField = $searchFieldType->getDefaultMatchField();
+        if (null === $searchFieldType) {
+            return $value;
         }
 
-        $fieldType = $definition[$matchField];
-        // convert core fieldtype search into a SPI/Search/Field to use in FieldValueMapper
-        $searchField = new SearchField('field', $value, $fieldType);
+        $searchField = new SearchField('field', $value, $searchFieldType);
+        $value = (array)$this->fieldValueMapper->map($searchField);
 
-        $value = $this->fieldValueMapper->map($searchField);
-
-        return current((array)$value);
+        return current($value);
     }
 }
--- a/lib/Query/Content/CriterionVisitor/Field.php
+++ b/lib/Query/Content/CriterionVisitor/Field.php
@@ -11,8 +11,11 @@
 namespace EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor;
 
 use EzSystems\EzPlatformSolrSearchEngine\Query\CriterionVisitor;
+use EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper;
 use eZ\Publish\Core\Search\Common\FieldNameResolver;
 use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
+use eZ\Publish\SPI\Search\Field as SearchField;
+use eZ\Publish\SPI\FieldType\Indexable;
 
 /**
  * Visits the Field criterion.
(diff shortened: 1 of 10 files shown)


pspanja:
Looks good, but it should be re-tested once https://github.com/ezsystems/ezpublish-kernel/pull/1655 is merged.

lib/Query/Content/CriterionVisitor/Field.php
```diff
@@ -27,36 +30,54 @@
...
+            $criterion->target
         );
     }
+
+    /**
+     * Map search field value to solr value using FieldValueMapper.
+     *
+     * @param mixed $value
+     *
+     * @param \eZ\Publish\SPI\Search\FieldType $searchFieldType
```
pspanja: Add empty line below and remove empty line above (separate return annotation and group param annotations).

lib/Query/Content/CriterionVisitor/Field.php
```diff
@@ -27,36 +30,54 @@
...
-     * Get field names.
+     * Get array of search fields.
      *
      * @param \eZ\Publish\API\Repository\Values\Content\Query\Criterion $criterion
-     * @param string $fieldDefinitionIdentifier
-     * @param string $fieldTypeIdentifier
-     * @param string $name
      *
-     * @return array
+     * @return array array of field types indexed by name.
```
pspanja: **An** array of field types indexed by name.
You can also hint this as `\eZ\Publish\SPI\Search\FieldType[]`.

lib/Query/Content/CriterionVisitor/Field.php
```diff
@@ -27,36 +30,54 @@
...
+    protected $fieldValueMapper;
+
+    /**
      * Create from content type handler and field registry.
      *
      * @param \eZ\Publish\Core\Search\Common\FieldNameResolver $fieldNameResolver
+     * @param \EzSystems\EzPlatformSolrSearchEngine\FieldValueMapper $fieldValueMapper
      */
-    public function __construct(FieldNameResolver $fieldNameResolver)
+    public function __construct(FieldNameResolver $fieldNameResolver, FieldValueMapper $fieldValueMapper)
```
pspanja: Please remove the descrtipion in the docblock as it is not relevant anymore.

tests/lib/Search/Query/CriterionVisitor/FullTextTest.php
```diff
@@ -100,8 +118,13 @@ public function testVisitBoostUnknownField()
 
     public function testVisitFuzzyBoost()
     {
-        $visitor = $this->getFullTextCriterionVisitor(array('title_1_s', 'title_2_s'));
-
+        $ftTextLine = new \eZ\Publish\Core\FieldType\TextLine\SearchField();
```
pspanja: This is not correct, it is search field type and not `Indexable` implementation.
`Fulltext` does not use it for mapping as it assumes string, but it still should be correct.

pspanja:
Looks ok, but it needs some description, remarks fixed and it needs to be re-tested after https://github.com/ezsystems/ezpublish-kernel/pull/1655 is merged.

andrerom:
> and it needs to be re-tested after ezsystems/ezpublish-kernel#1655 is merged.

that has been merged, and tests has been restarted ;)

lib/Query/Content/CriterionVisitor/Field.php
galileo: Removed

andrerom:
So, tagged a [new beta for kernel](https://github.com/ezsystems/ezpublish-kernel/releases/tag/v6.3.2-beta1) so we can bump version requirement here.

So please do the following changes to composer.json in this PR:
```diff
diff --git a/composer.json b/composer.json
index d97b98d..40f27b4 100644
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": "^5.4.4|~7.0",
-        "ezsystems/ezpublish-kernel": "~5.4.6@dev|~6.0.1@dev|^6.1.1@dev"
+        "ezsystems/ezpublish-kernel": "~5.4.7@dev|^6.3.2@dev"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.7",
@@ -27,7 +27,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.0.x-dev"
+            "dev-master": "1.1.x-dev"
         }
     }
 }
```

tests/lib/Search/Query/CriterionVisitor/FullTextTest.php
andrerom: @galileo when done with your current current spike on other task, could you fix this?

andrerom:
+1 as tests *(are about to)* pass now, but please fix the remark by @pspanja when you have time.

@pspanja any other remarks?

tests/lib/Search/Query/CriterionVisitor/FullTextTest.php
pspanja: I meant `FieldNameResolver::getFieldTypes()` that is here mocked was changed [here](https://github.com/ezsystems/ezpublish-kernel/pull/1655/files#diff-75f48ab676c919d25be0a06a14d42632R172) to return search field type (boolean, date, string etc) and not `Indexable` implementation for the field type.

`Fulltext` criterion [does not use it for casting](https://github.com/galileo/ezplatform-solr-search-engine/blob/5391bbaca495182380a807c050d559430b2b9f2d/lib/Query/Content/CriterionVisitor/FullText.php#L81), because it assumes search value is string. That is why the tests pass, but regardless, mock should still be correct here. Not really the topic here, but apparently we also miss `escapeQuote()` and `toString()` on that value.

andrerom: > I meant FieldNameResolver::getFieldTypes() that is here mocked was changed here to return search field type (boolean, date, string etc) and not Indexable implementation for the field type.

Roger

> but apparently we also miss escapeQuote() and toString() on that value.

where?

pspanja: Fulltext visitor, both main condition and boost by field:

* https://github.com/galileo/ezplatform-solr-search-engine/blob/5391bbaca495182380a807c050d559430b2b9f2d/lib/Query/Content/CriterionVisitor/FullText.php#L75
* https://github.com/galileo/ezplatform-solr-search-engine/blob/5391bbaca495182380a807c050d559430b2b9f2d/lib/Query/Content/CriterionVisitor/FullText.php#L82

andrerom:
Thanks @galileo

PR #46, opened 2016-05-25
EZP-25792: Deleted location don't affect subitems

galileo:
Issue: https://jira.ez.no/browse/EZP-25792

> Status: Ready for review
> Depends on https://github.com/ezsystems/ezpublish-kernel/pull/1658 *(integration tests)*

Explanation: 

The problem is that when we delete location we also want to delete all related sub location with related content, but if there is a content that have still another existing location that is out of the delete location subtree we just need to reindex this location instead of delete it.

Solution:

The solution is to find first which location need to be reindexed and with then need to be deleted with the content. As solr is handling location that are within content block we need to reindex or delete content index data.

### DIFF
--- a/lib/Gateway/Native.php
+++ b/lib/Gateway/Native.php
@@ -151,9 +151,7 @@ class Native extends Gateway
             $parameters['shards'] = $searchTargets;
         }
 
-        $queryString = $this->generateQueryString($parameters);
-
-        return $this->search($queryString, $parameters);
+        return $this->search($parameters);
     }
 
     public function searchAllEndpoints(Query $query)
@@ -165,9 +163,7 @@ class Native extends Gateway
             $parameters['shards'] = $searchTargets;
         }
 
-        $queryString = $this->generateQueryString($parameters);
-
-        return $this->search($queryString, $parameters);
+        return $this->search($parameters);
     }
 
     /**
@@ -540,12 +536,14 @@ class Native extends Gateway
     /**
      * Perform request to client to search for records with query string.
      *
-     * @param string $queryString
      * @param array $parameters
+     *
      * @return mixed
      */
-    protected function search($queryString, array $parameters)
+    protected function search(array $parameters)
     {
+        $queryString = $this->generateQueryString($parameters);
+
         $response = $this->client->request(
             'GET',
             $this->endpointRegistry->getEndpoint(
--- a/lib/Gateway/Native.php
+++ b/lib/Gateway/Native.php
@@ -173,6 +173,37 @@ class Native extends Gateway
         return $result;
     }
 
+    public function findAllSomething(Query $query)
+    {
+        $parameters = $this->contentQueryConverter->convert($query);
+
+        $searchTargets = $this->endpointResolver->getEndpoints();
+        if (!empty($searchTargets)) {
+            $parameters['shards'] = $searchTargets;
+        }
+
+        $queryString = $this->generateQueryString($parameters);
+
+        $response = $this->client->request(
+            'GET',
+            $this->endpointRegistry->getEndpoint(
+                $this->endpointResolver->getEntryEndpoint()
+            ),
+            "/select?{$queryString}"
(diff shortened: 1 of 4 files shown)


galileo:
I think I solved most issues for escaping and preparing query for deleted and update for solr documents.

So I have the good solr documnets in the solr engine but now when I use the second search just for an

Content\Id(14)

I recieve this kind of query:

~~~
(
  (
    document_type_id:content
  ) AND (
    (
      content_id:"14"
    ) AND (
      (
        section_id:"1"
      ) OR (
        (
          type_id:"25" 
          OR 
          type_id:"33"
        ) AND (
          section_id:"3"
        )
      )
    )
  ) AND (
    meta_indexed_is_main_translation_b:true
  )
)
~~~

Instead of this kind of query, used for the first search:

~~~

(
  (
    document_type_id:content
  ) AND (
    content_id:"14"
  ) AND (
    meta_indexed_is_main_translation_b:true
  )
)

~~~

So do anyone have any clues what is the reason? I just diagnosed it and any kind of information in this area will be very helpfull.

ping @pspanja

galileo:
I found what are those additional elements in my search query.

Because we are dealing there with users section we may destroy credentials if we touch this section, and I did this here.

So I will create new elements in the content area to be sure to test just the fixed behavior and not to deal with permssions.

Currently i have just this fixtures:

~~~
/1/2/ -> Home
/1/2/60/ -> Contact Us
~~~

What is the best approach to add here elements?
> For now I will just write all creating stuff to test this case in this test but maybe some fixtures would be better? 

ping @pspanja @andrerom

andrerom:
@galileo Not fully following the question, what did you have in mind here? What do you need to do? Are you asking how you can create content in integration tests?

galileo:
This question is actually related more to the publish-kernel because this is the place where our integration test lives. 

I need to prepare a whole new structure of content with items and subitems and with additional location.
So I done this here straight in the code on the test begin:

https://github.com/ezsystems/ezpublish-kernel/pull/1658/files#diff-2a6b4eff68887cf7238f0c51cb2ab44cR183

I create a container with sub items and for one of them I add multiple locations.

So the main question now is, can this stay like this or is there any other way to do those fixtures? 

@andrerom

galileo:
Why the final solution is different that the previous ?

Because when we need the `_root_` id if we want to delete whole `block` in solr. I didn't found any useable alternative to remove `block's`, so the flow is now to find all those content that need to be deleted and then remove them by `_root_` id.

Most risky places:

 - https://github.com/ezsystems/ezplatform-solr-search-engine/pull/46/files#diff-0f48e7179f8211995cf1fe2167d4e508R191
 - https://github.com/ezsystems/ezplatform-solr-search-engine/pull/46/files#diff-b212c6c9f6ec49c79e915504ddc0d4d9R56


@andrerom @pspanja

lib/Gateway/Native.php
```diff
@@ -153,24 +153,21 @@ protected function internalFind(array $parameters, array $languageSettings = arr
...
-            throw new RuntimeException(
-                '->response not set: ' . var_export(array($result, $parameters), true)
-            );
+        $searchTargets = $this->getAllSearchTargets();
+        if (!empty($searchTargets)) {
+            $parameters['shards'] = $searchTargets;
         }
 
-        return $result;
+        $queryString = $this->generateQueryString($parameters);
```
andrerom: Seems like we can move this line into `search()` so we only need to pass parameters?

lib/Handler.php
```diff
@@ -208,6 +208,8 @@ public function findLocations(LocationQuery $query, array $fieldFilters = array(
      * @param string[] $fieldPaths
      * @param int $limit
      * @param \eZ\Publish\API\Repository\Values\Content\Query\Criterion $filter
+     *
+     * @throws \Exception
```
andrerom: skip

andrerom:
+1

PR #47, opened 2016-06-07
EZP-25756: Can't index data in solrBundle

alongosz:
**Related JIRA issue**: https://jira.ez.no/browse/EZP-25756

It fixes crashing of _ezplatform:solr_create_index_ command in 2 cases:
- [x] [Non-existent content object version](https://github.com/alongosz/ezplatform-solr-search-engine/blob/bc5b5db96e0e15e9f7e55fc54496ec4073b0ca3d/bundle/Command/SolrCreateIndexCommand.php#L90-L98) ([also logs a warning](https://github.com/alongosz/ezplatform-solr-search-engine/blob/bc5b5db96e0e15e9f7e55fc54496ec4073b0ca3d/bundle/Command/SolrCreateIndexCommand.php#L115))
- [x] Stateless content object ([here](https://github.com/alongosz/ezplatform-solr-search-engine/blob/bc5b5db96e0e15e9f7e55fc54496ec4073b0ca3d/lib/Handler.php#L249-L253) unable to log this unfortunately)

### DIFF
--- a/bundle/Command/SolrCreateIndexCommand.php
+++ b/bundle/Command/SolrCreateIndexCommand.php
@@ -15,6 +15,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Input\InputArgument;
 use eZ\Publish\SPI\Persistence\Content\ContentInfo;
+use eZ\Publish\Core\Base\Exceptions\NotFoundException;
 use EzSystems\EzPlatformSolrSearchEngine\Handler as SolrSearchEngineHandler;
 use RuntimeException;
 use PDO;
@@ -78,6 +79,8 @@ EOT
         $progress = $this->getHelperSet()->get('progress');
         $progress->start($output, $totalCount);
         $i = 0;
+        $logger = $this->getContainer()->get('logger');
+        /* @var \Psr\Log\LoggerInterface $logger */
         do {
             $contentObjects = array();
 
@@ -85,11 +88,17 @@ EOT
                 if (!$row = $stmt->fetch(PDO::FETCH_ASSOC)) {
                     break;
                 }
-
-                $contentObjects[] = $persistenceHandler->contentHandler()->load(
-                    $row['id'],
-                    $row['current_version']
-                );
+                try {
+                    $contentObjects[] = $persistenceHandler->contentHandler()->load(
+                        $row['id'],
+                        $row['current_version']
+                    );
+                } catch (NotFoundException $e) {
+                    $progress->clear();
+                    $output->write("\r"); // get rid of padding (side effect of displaying progress bar)
+                    $logger->warning("Could not load current version of Content with id ${row['id']}, so skipped for indexing. Full exception: " . $e->getMessage());
+                    $progress->display();
+                }
             }
 
             if (!empty($contentObjects)) {
--- a/lib/Handler.php
+++ b/lib/Handler.php
@@ -246,7 +246,11 @@ class Handler implements SearchHandlerInterface
         $documents = array();
 
         foreach ($contentObjects as $content) {
-            $documents[] = $this->mapper->mapContentBlock($content);
+            try {
+                $documents[] = $this->mapper->mapContentBlock($content);
+            } catch (NotFoundException $ex) {
+                // ignore content objects without assigned state id
+            }
         }
 
         if (!empty($documents)) {


alongosz:
Ready for review, ping @andrerom

lib/Handler.php
```diff
@@ -246,7 +246,11 @@ public function bulkIndexContent(array $contentObjects)
         $documents = array();
 
         foreach ($contentObjects as $content) {
-            $documents[] = $this->mapper->mapContentBlock($content);
+            try {
+                $documents[] = $this->mapper->mapContentBlock($content);
+            } catch (NotFoundException $ex) {
+                // ignore stateless content
```
andrerom: what do you mean by stateless?

bundle/Command/SolrCreateIndexCommand.php
```diff
@@ -78,18 +79,23 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-                    $row['current_version']
-                );
+                try {
+                    $contentObjects[] = $persistenceHandler->contentHandler()->load(
+                        $row['id'],
+                        $row['current_version']
+                    );
+                } catch (NotFoundException $e) {
+                    // ignore content without versions
+                    array_push($warnings, $e->getMessage());
```
andrerom: Typically we rather do:
`$warnings[] = $e->getMessage();`

andrerom: But warning message here will be to generic, it will say object x with version y could not be loaded, better if we create a specific warning message .

andrerom:
Looks ok, but wondering if using the actual logger would be better.
Would that also break progress bar?

pov/input @bdunogier ?

lib/Handler.php
alongosz: When a content object does not have corresponding entry in the database table _ezcobj_state_link_ a [NotFoundException is thrown](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/Persistence/Legacy/Content/ObjectState/Handler.php#L339). In ezstudio-demo data I found such a case where object had a version but didn't have state assigned. Would it be enough if I changed this comment to "ignore content objects without assigned state id"?
ping @andrerom

bundle/Command/SolrCreateIndexCommand.php
alongosz: > But warning message here will be to generic, it will say object x with version y could not be loaded, better if we create a specific warning message .

Indeed, now we have e.g.: "Could not find 'content' with identifier 'contentId: 101, versionNo: 1'"
@andrerom: Do you have in mind something more user-friendly, like "Unable to find any version of the Content Object Id = 101"?

alongosz:
> Looks ok, but wondering if using the actual logger would be better.
Would that also break progress bar?

Unfortunately it looks like this:

```
Indexing Content...
42/49 [========================>---]  85%[2016-06-07 14:59:53] app.WARNING: Unable to find any version of the Content object with ID 101
49/49 [============================] 100%
```

so it still would require displaying it after progress bar finishes. I tried to find a better way to handle ProgressHelper and logger but came up empty.

bundle/Command/SolrCreateIndexCommand.php
andrerom: `$warnings[] = "Could not load current version of Content with id ${row['id']}, full exception: " . $e->getMessage();

lib/Handler.php
andrerom: that would be ok

But is this common? Afaik it should not be as I think Repository tried to always create states on content creation, but I might remember it wrong.

If I remember it right, then this should be a warning as well:
`$warnings[] = "Could not pull in meta data about Content with id ${row['id']}, so it is not indexed. full exception: " . $e->getMessage();`

pspanja: IMHO this should be done in the command and not here.

andrerom: agreed, forgot this was not in command.

@alongosz then just update the comment like you proposed for now.

alongosz:
I've corrected one comment and warning message. In command class a logger is used to display warnings
Ready for final review @andrerom , @pspanja

andrerom:
+1

bdunogier:
+1, but would be better if what @andrerom found worked.

alongosz:
Well, I assumed `clear()` and `display()` would repeat entire progress bar after displaying a message, so I didn't check it 2 days ago while investigating - my mistake :) 

Actually it works almost as expected:

```
Indexing Content...
                                          [2016-06-09 10:26:08] app.WARNING:  Could not load current version of Content with id 101, so skipped for indexing. Full exception: Could not find 'content' with identifier 'contentId: 101, versionNo: 1'  
 49/49 [============================] 100%
```

Notice padding of a message. A workaround would be to call `$output->write("\r");` after `$progress->clear()` and before `$logger->warning`. However I don't know if it will work on Mac :)

With this workaround I was able to get desired output:
```
Indexing Content...
[2016-06-09 10:42:11] app.WARNING: Could not load current version of Content with id 101, so skipped for indexing. Full exception: Could not find 'content' with identifier 'contentId: 101, versionNo: 1'  
 49/49 [============================] 100%
```
I checked this for more than one warning - progress bar remains at the end of an output :)

Can I change it like this @andrerom , @bdunogier ?

andrerom:
yes please :)

lib/Handler.php
```diff
@@ -246,7 +246,11 @@ public function bulkIndexContent(array $contentObjects)
         $documents = array();
 
         foreach ($contentObjects as $content) {
-            $documents[] = $this->mapper->mapContentBlock($content);
+            try {
+                $documents[] = $this->mapper->mapContentBlock($content);
+            } catch (NotFoundException $ex) {
+                // ignore content objects without assigned state id
```
andrerom: SIDE NOTE: We should ideally figure out how to deal with warnings in a good way for output before we merge change to `SPI\Search\Handler` interface for adding `bulkIndexContent`, as is being done in WordIndexer kernel PR.

Most likely solution:
- `function generateDocumentByContent( Content $content ) : Document`
- `function bulkIndexDocuments( Document[] $documents ) : void`
- Hence moving the part where you might get exception to indexer command so we can log warnings

/cc @pspanja @alongosz

andrerom:
🎉

PR #53, opened 2016-06-30
Bump dependencies for 1.1 branch

pspanja:


pspanja:
As we discussed upcoming changes for v1.1 and kernel 6.5 ( branch already aliased there), I asked @bdunogier how this should look like.

Ping @andrerom

composer.json
```diff
@@ -10,8 +10,8 @@
         }
     ],
     "require": {
-        "php": "^5.4.4|~7.0",
-        "ezsystems/ezpublish-kernel": "~5.4.7@dev|~6.0.1.3|~6.1.2@dev|^6.3.2@dev"
+        "php": "~5.5|~7.0",
+        "ezsystems/ezpublish-kernel": "^6.5"
```
andrerom: might have to be 6.5@dev

pspanja: Indeed, updated.

bdunogier:
+1

But please use a slightly more informative commit message. I usually go for "[composer] updated dependencies for 1.1"

andrerom:
+1

pspanja:
Thanks, merging when Travis confirms (will rebase as suggested).

PR #57, opened 2016-08-01
Active spellcheck suggestion in solr bundle

mathieupoulenard:
Active spell check suggestion in Solr bundle.

I also add few lines in schema.xml 

``` xml
    <field name="spelling" type="text" indexed="true" multiValued="true" stored="false"/>
    <copyField source="*_fulltext" dest="spelling" />
```

and few lines in solrconfig.xml

``` xml
 <requestHandler name="/select" class="solr.SearchHandler" startup="lazy">
    <lst name="defaults">
      <bool name="distrib">true</bool>
      <!-- spell check component configuration -->
      <str name="spellcheck">true</str>
      <str name="spellcheck.count">5</str>
      <str name="spellcheck.collate">true</str>
      <str name="spellcheck.maxCollationTries">5</str>
      <str name="spellcheck.alternativeTermCount">5</str>
    </lst>
    <arr name="last-components">
      <str>spellcheck</str>
    </arr>
  </requestHandler>

  <searchComponent name="spellcheck" class="solr.SpellCheckComponent">
    <lst name="spellchecker">
      <!-- decide between dictionary based vs index based spelling suggestion, in most cases it makes sense to use index based spell checker as it only generates terms which are actually present in your search corpus -->
      <str name="classname">solr.IndexBasedSpellChecker</str>
      <!-- field to use -->
      <str name="field">spelling</str>
      <!-- buildOnCommit|buildOnOptimize -->
      <str name="buildOnCommit">true</str>
      <!-- $solr.solr.home/data/spellchecker-->
      <str name="spellcheckIndexDir">./spellchecker</str>
      <str name="accuracy">0.7</str>
      <float name="thresholdTokenFrequency">.0001</float>
    </lst>
  </searchComponent>
```

### DIFF
--- a/lib/Query/Common/QueryConverter/NativeQueryConverter.php
+++ b/lib/Query/Common/QueryConverter/NativeQueryConverter.php
@@ -63,7 +63,8 @@ class NativeQueryConverter extends QueryConverter
     {
         $params = array(
             'defType' => 'edismax',
-            'q.alt' => $this->criterionVisitor->visit($query->query),
+            'q' => $this->criterionVisitor->visit($query->query),
+            'q.alt' => '*:*',
             'fq' => $this->criterionVisitor->visit($query->filter),
             'sort' => $this->getSortClauses($query->sortClauses),
             'start' => $query->offset,
--- a/lib/Query/Common/QueryConverter/NativeQueryConverter.php
+++ b/lib/Query/Common/QueryConverter/NativeQueryConverter.php
@@ -73,7 +73,7 @@ class NativeQueryConverter extends QueryConverter
         );
 
         // use solr param q when search with fulltext query value
-        if (is_null($query->query->value)) {
+        if (null === $query->query->value) {
             $params['q.alt'] = $this->criterionVisitor->visit($query->query);
         } else {
             // solr q param enable suggestions in response
--- a/lib/Query/Common/QueryConverter/NativeQueryConverter.php
+++ b/lib/Query/Common/QueryConverter/NativeQueryConverter.php
@@ -63,8 +63,7 @@ class NativeQueryConverter extends QueryConverter
     {
         $params = array(
             'defType' => 'edismax',
-            'q' => $this->criterionVisitor->visit($query->query),
-            'q.alt' => '*:*',
+            'q.alt' => $this->criterionVisitor->visit($query->query),
             'fq' => $this->criterionVisitor->visit($query->filter),
             'sort' => $this->getSortClauses($query->sortClauses),
             'start' => $query->offset,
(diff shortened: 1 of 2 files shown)


lib/Query/Common/QueryConverter/NativeQueryConverter.php
```diff
@@ -72,6 +72,14 @@ public function convert(Query $query)
             'wt' => 'json',
         );
 
+        // use solr param q when search with fulltext query value
+        if (is_null($query->query->value)) {
```
andrerom: `=== null`

andrerom:
wdyt @pspanja ?

lib/Query/Common/QueryConverter/NativeQueryConverter.php
```diff
@@ -72,6 +72,14 @@ public function convert(Query $query)
             'wt' => 'json',
         );
 
+        // use solr param q when search with fulltext query value
+        if (null === $query->query->value) {
```
pspanja: Is this intended to recognize `FullText` criterion?

mathieupoulenard: Yes, il no value in FullText criterion we use q.alt and if we have a user query we use q to have solr suggestions in response

pspanja: The problem with that is that `FullText` could be used in `Query::$filter` or in some logical structure, so this can't work reliably. That actually means we can't extract a single search term for `q` parameter, unless we limit `Query::$query` to `FullText` criterion.

ale42: We can use `FullText` in `Query::$filter`, this will not change anything because we only check that `$query->query->value` is _not null_, so we will use `q.alt` as it's actually done. Value of `filter` is not checked here. 
Maybe instead of checking if `$query->query->value` is _not null_, we should check if we have a `fulltext` criterion?

andrerom:
Todo:
- [ ] Seems to not return data in the way [specified in the API](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/API/Repository/Values/Content/Search/SearchResult.php#L35-L40), that said the API doc might be too complex here, this is intended to explored during implementation _(this being the first attempt)_
- [ ] I think it is ok that spellcheck, just like scoring, only works on `->query`, however open question is 1. `q` vs `q.alt` 2. Does this imply only one criterion, namely FullText, will be allowed on `->query`, if so that is a 2.x API break.

pspanja:
Thanks for this guys, it finally got it's turn :)
Closing in favor of https://github.com/ezsystems/ezplatform-solr-search-engine/pull/81

mathieupoulenard:
Cool !

PR #63, opened 2016-09-07
EZP-26141: Search appears to conduct some form of 'OR' search not and 'AND' search

pspanja:
Resolves https://jira.ez.no/browse/EZP-26141

### TODO
- [x] add tests to kernel repo

### DIFF
--- a/tests/lib/Search/Query/CriterionVisitor/FullTextTest.php
+++ b/tests/lib/Search/Query/CriterionVisitor/FullTextTest.php
@@ -10,14 +10,16 @@
  */
 namespace EzSystems\EzPlatformSolrSearchEngine\Tests\Search\Query\CriterionVisitor;
 
-use EzSystems\EzPlatformSolrSearchEngine\Tests\Search\TestCase;
 use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
+use eZ\Publish\Core\FieldType\TextLine\SearchField;
+use eZ\Publish\SPI\Search\FieldType\StringField;
+use EzSystems\EzPlatformSolrSearchEngine\Tests\Search\TestCase;
 use EzSystems\EzPlatformSolrSearchEngine\Query;
 
 /**
  * Test case for FullText criterion visitor.
  *
- * @covers \EzSystems\EzPlatformSolrSearchEngine\Content\CriterionVisitor\FullText
+ * @covers \EzSystems\EzPlatformSolrSearchEngine\Query\Content\CriterionVisitor\FullText
  */
 class FullTextTest extends TestCase
 {
@@ -69,6 +71,18 @@ class FullTextTest extends TestCase
         );
     }
 
+    public function testVisitSimpleMultipleWords()
+    {
+        $visitor = $this->getFullTextCriterionVisitor();
+
+        $criterion = new Criterion\FullText('Hello World');
+
+        $this->assertEquals(
+            '(text:(Hello World))',
+            $visitor->visit($criterion)
+        );
+    }
+
     public function testVisitFuzzy()
     {
         $visitor = $this->getFullTextCriterionVisitor();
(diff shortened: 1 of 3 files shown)


andrerom:
So what is supported here now?

Is this supported?
- `Wild dog` will hit on content with both
-- and rank those with exact phrase highest
- `"Wild dog"` will hit on exact phrases
- Any kind of `AND`, `OR`, `+`, `-` 


Just asking to get a clear picture, cause would be cool to support some or all of above , even if we don't support it on legacy search yet.

pspanja:
For now I went for words and implicit `OR` operator, since that is how Legacy engine behaves (`Fulltext` criterion actually documents only implicit `AND`). But if we say Legacy engine will behave differently, we can go for all you described, both for Solr and Elastic, plus even grouping with parenthesis.

Only problem there is that we need to allow targeting specific fields for boosting, so it will require some parsing to escape everything properly. Also, fuzziness can't be applied to phrases/groups, only words, so we need to decide how to handle that case.

andrerom:
Ok, seems clear, if possible lets do the following for now to improve it slightly:
- if possible boost the whole phrase so user will get search results in following order:
 - Exact hits on `"Red Apple"`
 - Hits that contains both words
 - Hits that contains both works with fuzziness
 - Hits that contain one of them
 - Hits that contain one of them with fuzziness

andrerom:
Note: We should apply this to 1.0 branch as well.

pspanja:
Added exact phrase match (on boosted fields as well) in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/63/commits/2b9627631a5c36bd98fcc8cd3d3816705a22b63b

We will probably need to change this when introducing operators as phrase won't be exactly defined in that case.

pspanja:
Added more tests that show what is generated in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/63/commits/cfb582d066f6f7e60e4e83cc69be20e809fac7da

tests/lib/Search/Query/CriterionVisitor/FullTextTest.php
```diff
@@ -64,7 +66,19 @@ public function testVisitSimple()
...
+    }
+
+    public function testVisitSimpleMultipleWords()
+    {
+        $visitor = $this->getFullTextCriterionVisitor();
+
+        $criterion = new Criterion\FullText('Hello World');
+
+        $this->assertEquals(
+            '(text:(Hello World) OR text:"Hello World")',
```
andrerom: ok so `text:(Hello World)` was given to Solr all along, thought it was `text:(Hello) OR text:(World)`, so then adding the phrasing provably does not make a difference then. Solr is potentially already ranking exact hits first, or?

pspanja: Actually, `text:(Hello World)` is functionally identical to`text:(Hello) OR text:(World)`.
So matching phrase does make a difference, for example, for the added [test in kernel](https://github.com/ezsystems/ezpublish-kernel/pull/1770/files#diff-dccc5965065be67bd0ff884075ccf5bdR4333), the order is same, but score difference is more pronounced. Without phrase match:

1. ID:60, 0.9023705
2. ID:62, 0.73013973
3. ID:61, 0.1964612

And with phrase match:

1. ID:60, 2.6263494
2. ID:62, 0.95123565
3. ID:61, 0.04279712

But I'm not sure if we really need this. Would rather go with simpler approach and add proper explicit phrase matching later on.

andrerom: Ok, then I would likte us to keep it, the order might be the same here, but in a database with thousands of content items this will make a huge differance until we replace it with a proper approach _(potentially across engines even)_.

andrerom: So assuming ID 60 will _not_ anymore score highest with "World Hello", then I agree we delete it.

pspanja: Indeed it can mess with boost, but not on that example. Better example, indexing:

1. ID:60, name:'red', sort_name:'red apple'
2. ID:61: name:'apple red', short_name:'one'

Then, with phrase matching the result for 'red apple' is:

1. ID:60, 2.4462724
2. ID:61: 0.15659831

And for 'apple red' the order is changed:

1. ID:61, 0.750654
2. ID:60: 0.15101993

Without phrase matching order and score are kept:

1. ID:60, 0.8453717
2. ID:61: 0.7003289

I'll remove phrase matching.

pspanja: Rebased to previous version.

andrerom:
+1

alongosz:
+1

andrerom:
Ok, ready to be merged into 1.0 and master 👌🏻

pspanja:
Closing in favor of https://github.com/ezsystems/ezplatform-solr-search-engine/pull/64

PR #66, opened 2016-09-14
Fix EZP-26307: incorrect bulk_count results in index loop

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-26307

Problem: 
`bulk_count` parameter is cast to int without validation (converted to 0), while the loop would effectively index `$bulkCount+1` contents per run.
This would result in an infinite loop.

Fix by validating the parameter and indexing the correct number of contents.


andrerom:
+1, but you should rebase this on 1.0 branch and re target PR.

joaoinacio:
> +1, but you should rebase this on 1.0 branch and re target PR.

done.

alongosz:
+1

andrerom:
Merging, thanks @joaoinacio

PR #67, opened 2016-09-27
EZP-26368: As a developer I want to index custom data with Solr Search Engine

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-26368

This is https://github.com/ezsystems/ezplatform-solr-search-engine/pull/42 continued.
As was discussed previously `LocationTranslation` mapper is here removed, to be proposed in a separate PR.

This implements document field mappers, aka indexing plugins feature.
Three types of base plugins are implemented:
1. `Content::mapFields(Content $content)`
2. `ContentTranslation::mapFields(Content $content, $languageCode)`
3. `Location::mapFields(Location $location)`

For each of these an `Aggregate` plugin is implemented and used (6 places total):
1. `Content::mapFields(Content $content)`
   1. Block - indexes fields on the block document level -- meaning to each Content and Location document in the block
   2. Content - indexes fields to each Content document
2. `ContentTranslation::mapFields(Content $content, $languageCode)`
   1. BlockTranslation - indexes fields on the block level per translation
   2. ContentTranslation - indexes fields to Content document per translation
3. `Location::mapFields(Location $location)`
   1. Location - indexes field to each Location document

The same, but starting from extension points (service tags):
1. Block: `ezpublish.search.solr.document_field_mapper.block`
   - Accepts `Content::mapFields(Content $content)`
2. Block translation: `ezpublish.search.solr.document_field_mapper.block_translation`
   - Accepts `ContentTranslation::mapFields(Content $content, $languageCode)`
3. Content: `ezpublish.search.solr.document_field_mapper.content`
   - Accepts `Content::mapFields(Content $content)`
4. Content translation: `ezpublish.search.solr.document_field_mapper.content_translation`
   - Accepts `ContentTranslation::mapFields(Content $content, $languageCode)`
5. Location: `ezpublish.search.solr.document_field_mapper.location`
   - Accepts `Location::mapFields(Location $location)`

Concrete plugins are registered to `Aggregate` plugin in a container compiler pass, using service tags.
At the moment no concrete plugins are implemented, followup to this would be refactoring `DocumentMapper` to use only plugins to map fields. How that would look like can be seen here: https://github.com/pspanja/ezplatform-solr-search-engine/pull/1
This is `DocumentMapper` fully refactored to use plugins: https://github.com/pspanja/ezplatform-solr-search-engine/blob/3a439bcf098b7e5a6c397204fc5849737cb1a60b/lib/DocumentMapper/NativeDocumentMapper.php


pspanja:
Ready for review, ping @andrerom @alongosz @bdunogier

lib/Container/Compiler/BaseDocumentFieldMapperPass.php
```diff
@@ -0,0 +1,50 @@
...
+namespace EzSystems\EzPlatformSolrSearchEngine\Container\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Base compiler pass for aggregate document field mappers.
+ */
+abstract class BaseDocumentFieldMapperPass implements CompilerPassInterface
```
bdunogier: Not a real review (I don't _request_ the change), but we have similar compiler passes over the place. Shouldn't we try to have a common one that we can re-use in all of our packages ?

lib/DocumentMapper/FieldMapper.php
```diff
@@ -0,0 +1,18 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
+ */
+namespace EzSystems\EzPlatformSolrSearchEngine\DocumentMapper;
+
+/**
+ * Base class for document field mappers.
+ */
+abstract class FieldMapper
```
bdunogier: Why not an interface rather than an abstract class ? i'd actually really suggest an interface. An abstract class would have a particular behaviour, while `FieldMapper` hints at a "role" (e.g. what it does).

lib/DocumentMapper/FieldMapper/Content.php
```diff
@@ -0,0 +1,40 @@
...
+
+use EzSystems\EzPlatformSolrSearchEngine\DocumentMapper\FieldMapper;
+use eZ\Publish\SPI\Persistence\Content as SPIContent;
+
+/**
+ * Base class for Content document field mapper.
+ *
+ * Content document field mapper maps Content to the search fields for Content document.
+ */
+abstract class Content extends FieldMapper
```
bdunogier: Same as above (interface vs. abstract class).

Also, I tend to avoid naming things "Content" or similar. We end up with an s'load of those. `ContentFieldMapper` is still short enough, is it not ?

(I won't be fanatic about it, just an improvement imho).

lib/DocumentMapper/FieldMapper/Location/Aggregate.php
```diff
@@ -0,0 +1,66 @@
...
+ */
+namespace EzSystems\EzPlatformSolrSearchEngine\DocumentMapper\FieldMapper\Location;
+
+use EzSystems\EzPlatformSolrSearchEngine\DocumentMapper\FieldMapper\Location as LocationMapper;
+use eZ\Publish\SPI\Persistence\Content\Location;
+
+/**
+ * Aggregate implementation of Location document field mapper.
+ */
+class Aggregate extends LocationMapper
```
bdunogier: Is it me or all of your Aggregate mappers have the same code ?

lib/DocumentMapper/NativeDocumentMapper.php
```diff
@@ -703,4 +784,121 @@ protected function getObjectStateIds($contentId)
...
+     * @param \eZ\Publish\SPI\Persistence\Content\Location $location
+     * @param string $languageCode
+     *
+     * @return \eZ\Publish\SPI\Search\Field[]
+     */
+    private function getLocationTranslationFields(Location $location, $languageCode)
+    {
+        $fields = [];
+
+        if ($this->locationTranslationFieldMapper->accept($location, $languageCode)) {
```
bdunogier: Would be shorter with a return early:

```
if ($this->locationTranslationFieldMapper->accept($location, $languageCode)) {
  return [];
}

return $this->locationTranslationFieldMapper->mapFields($location, $languageCode);

File: lib/DocumentMapper/NativeDocumentMapper.php [hunk 6]
```diff
@@ -83,6 +117,12 @@ class NativeDocumentMapper implements DocumentMapper
     /**
      * Creates a new document mapper.
      *
+     * @param \EzSystems\EzPlatformSolrSearchEngine\DocumentMapper\FieldMapper\Content $blockFieldMapper
```
bdunogier: This is sparta

Jokes aside, could this be refactored to be a little slimer ? :)

pspanja:
@alongosz 
> (I would remove @version //autogentag// from all new files, beside that it looks good)

Removed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/67/commits/a2b3a89142a19917ae1032a5d84c15314e533388

pspanja:
@andrerom 
> do you also have updated PR with the followup changes where inline logic is removed in favor of using this plugin system?

Yes, linked to it from description: https://github.com/pspanja/ezplatform-solr-search-engine/pull/1
`DocumentMapper` should eventually look like this: https://github.com/pspanja/ezplatform-solr-search-engine/blob/3a439bcf098b7e5a6c397204fc5849737cb1a60b/lib/DocumentMapper/NativeDocumentMapper.php

lib/Container/Compiler/BaseDocumentFieldMapperPass.php
pspanja: I guess we could do that, it should probably be in `ezpublish-kernel` repo.

lib/DocumentMapper/FieldMapper/Location/Aggregate.php
pspanja: It's you :) They are mostly the same, but accept different mapper type.

lib/DocumentMapper/NativeDocumentMapper.php
pspanja: The only possibility I see is to move it up, on the same level as `DocumentMapper`.
I actually like that idea, it's the same approach I took in https://github.com/pspanja/ezplatform-solr-search-engine/pull/4 with a number of other services.

lib/DocumentMapper/NativeDocumentMapper.php
pspanja: I would, but I would not be able to process it at a glance :)
It's really a matter of style.

lib/DocumentMapper/FieldMapper.php
pspanja: Actually I think I'll just remove this, as it serves not purpose ATM. We can introduce it later on if need be.

pspanja: Removed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/67/commits/a433031df86b056143b04a2d0173298079289f1b

lib/DocumentMapper/FieldMapper/Content.php
pspanja: Not sure about the benefit of the interface here, it's perfectly fine for abstract class not to have it's own implementation. Role would be more like `FieldMapping`.

Will rename when moving up.

pspanja: Renamed (and moved up) in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/67/commits/30ffce170f623e0c284b70e4a0b3a5889ebf3acf.

lib/DocumentMapper/NativeDocumentMapper.php
pspanja: Moved up in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/67/commits/30ffce170f623e0c284b70e4a0b3a5889ebf3acf

pspanja:
I think I addressed all remarks.
I changed the structure a bit, moving `FieldMapper` up one level, so please take another look.

andrerom:
What do you think @bdunogier? From extensibility perspective do we hit a good spot here?

pspanja:
There were no significant commits to `master` since the last change here, so with @bdunogier's approval I can move this forward.

bdunogier:
Is there an example that I can look at somewhere ? It would help a lot :)

pspanja:
The one linked in the description, should be virtually identical, aside from placement, removal of `LocationTranslation` mapper and removal of the common abstract class. The code you see there is basically what we'll be going for next with `DocumentMapper` refactoring to use plugins: 

https://github.com/pspanja/ezplatform-solr-search-engine/tree/3a439bcf098b7e5a6c397204fc5849737cb1a60b/lib/DocumentMapper

bdunogier:
> aside from placement, removal of LocationTranslation mapper and removal of the common abstract class

"lol" ;)

pspanja:
You're welcome. This is the code with consolidated field names and visitors, so basically removing unnecessary duplication even further:

https://github.com/pspanja/ezplatform-solr-search-engine/tree/3d7db66c36771fb26b5d70a94fe6fa7f496accd5/lib/DocumentMapper

Also do check the container configuration:

https://github.com/pspanja/ezplatform-solr-search-engine/blob/3d7db66c36771fb26b5d70a94fe6fa7f496accd5/lib/Resources/config/container/solr/document_field_mappers.yml

lib/Container/Compiler/FieldMapperPass/ContentFieldMapperPass.php
```diff
@@ -0,0 +1,20 @@
...
+namespace EzSystems\EzPlatformSolrSearchEngine\Container\Compiler\FieldMapperPass;
+
+use EzSystems\EzPlatformSolrSearchEngine\Container\Compiler\BaseFieldMapperPass;
+
+/**
+ * Compiler pass for aggregate document field mapper for the Content document.
+ */
+class ContentFieldMapperPass extends BaseFieldMapperPass
+{
+    const AGGREGATE_MAPPER_SERVICE_ID = 'ezpublish.search.solr.field_mapper.content';
```
bdunogier: A bit late, so take that as an FYI, but you could have use one concrete class (`FieldMapperPass`), with the service id / service tag as arguments, and instantiate the Pass N times from the bundle, with the respective arguments. Less files, same result, easier to test.

bdunogier:
Looks okay from an extensibility pov. It would be much easier to review this aspect with documentation about the said extensibility (I've found out that writing this doc together with the feature really helps).

doc/specifications/document_field_mappers.md
```diff
@@ -0,0 +1,100 @@
+# Document Field Mappers
+
+As a website developer you will often find the need to index some additional data in the search
```
andrerom: website developer => developer

MarioBlazek:
Excellent, thanks @pspanja .

pspanja:
@andrerom 

> +1, but this implies master should be bumped to 1.2.x, so when this is merged, commit before should be branched of to 1.1 in case we need to do another 1.1 release.

A little bit late with this, now created `1.1`, alias bumped in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/73

PR #68, opened 2016-09-27
EZP-26098: Implement Search Indexer Service

alongosz:
> Related to: [EZP-26098](https://jira.ez.no/browse/EZP-26098).

This PR aligns with the most recent approach presented in the [Kernel PR #1738](https://github.com/ezsystems/ezpublish-kernel/pull/1738) implementing Search Engine Indexer Service for Solr, based on current version of `\EzSystems\EzPlatformSolrSearchEngineBundle\Command\SolrCreateIndexCommand`.

**TODO**:
- [x] Implement Search Engine Indexer Service for Solr (7ddf875).
- [x] Wait for [Kernel PR #1738](https://github.com/ezsystems/ezpublish-kernel/pull/1738) to be merged.

### DIFF
--- a/lib/Indexer.php
+++ b/lib/Indexer.php
@@ -1,0 +1,91 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformSolrSearchEngine;
+
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
+use eZ\Publish\Core\Search\Common\Indexer as SearchIndexer;
+use EzSystems\EzPlatformSolrSearchEngine\Handler as SearchHandler;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Output\OutputInterface;
+use PDO;
+use RuntimeException;
+
+class Indexer extends SearchIndexer
+{
+    /**
+     * @var \EzSystems\EzPlatformSolrSearchEngine\Handler
+     */
+    protected $searchHandler;
+
+    /**
+     * Create search engine index.
+     *
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     * @param int $iterationCount
+     * @param bool $commit
+     */
+    public function createSearchIndex(OutputInterface $output, $iterationCount, $commit)
+    {
+        $output->writeln('Creating Solr Search Engine Index...');
+
+        if (!$this->searchHandler instanceof SearchHandler) {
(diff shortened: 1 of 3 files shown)


lib/Indexer.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Create search engine index.
+     *
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     * @param int $iterationCount
+     * @param bool $commit
+     */
+    public function createSearchIndex(OutputInterface $output, $iterationCount, $commit)
+    {
+        $this->logger->notice('Creating Solr Search Engine Index...');
```
alongosz: So far, in commands, we've used directly `$output->writeln()` for such messages. I used here logger because this makes more sense to me - IMHO such event should be logged. `Notice` is documented as `Normal but significant events`, so it seems like the right choice.
However one problem with this is that by default it won't be displayed on `stdout`. It would either require `-v` switch passed to command or setting logger `VERBOSITY_NORMAL` to `NOTICE` (for Monolog [by default](http://symfony.com/doc/current/reference/configuration/monolog.html#full-default-configuration) it's `WARNING`).

@andrerom, @pspanja any POV on that?

andrerom:
Merged in https://github.com/ezsystems/ezpublish-kernel/commit/0a0887d2f2fad8c0d3a61eb53a50976d4c2c214d

andrerom:
Ping @pspanja

lib/Indexer.php
pspanja: Good point, I think we should use `$output->writeln()` for this.

pspanja:
Seems we have a conflict with `composer.json`. Aside from that and one other minor remark, this looks good to me.

lib/Indexer.php
alongosz: Ok, fixed.

alongosz:
I rebased this PR just now against `master`, squashing fixups and removing TMP Kernel branch.
For CI tests to pass, we have to wait for [Kernel PR #1738](https://github.com/ezsystems/ezpublish-kernel/pull/1738) to be merged.

lib/Resources/config/container/solr.yml
```diff
@@ -110,3 +111,14 @@ services:
...
+
+    ezpublish.spi.search.solr.indexer:
+        class: "%ezpublish.spi.search.solr.indexer.class%"
+        arguments:
+            - "@logger"
+            - "@ezpublish.api.storage_engine"
+            - "@ezpublish.api.storage_engine.legacy.dbhandler"
+            - "@ezpublish.spi.search.solr"
+        tags:
+            - {name: ezpublish.searchEngineIndexer, alias: solr}
```
andrerom: Missed this, but we should not use ezpublish in new tags we make right? /cc @bdunogier

andrerom: Ignore, we stay with the tag name as all follow this form.

andrerom:
Kernel PR merged, restarted Travis.

andrerom:
Merged, should pass now that php version on .travis.yml has been bumped.

PR #69, opened 2016-10-17
EZP-26756: FieldRelation location criterion visitor

florianalexandre:
JIRA issue: https://jira.ez.no/browse/EZP-26756

Allow the use of FieldRelation criterion in location queries
Use existing FieldRelation content criterion visitor

### DIFF
--- a/lib/Query/Location/CriterionVisitor/Field/FieldRelation.php
+++ b/lib/Query/Location/CriterionVisitor/Field/FieldRelation.php
@@ -1,0 +1,20 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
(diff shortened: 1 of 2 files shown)


andrerom:
Thanks for the contribution so far, hopefully we'll get around to have a look at this next week when @pspanja is back to give you some review feedback :)

pspanja:
This looks good, but we should also have integration tests for it in https://github.com/ezsystems/ezpublish-kernel.
We already have tests for `FieldRelation` with Content search, they would only need to be adapted to work for Location search as well, in the same style as we already do with other tests.

@florianalexandre let me know if you are up for this.

We'll need an issue as well :)

andrerom:
Status for those wondering _(including myself)_:
- [x] visitor _(fixed in master, aka 1.2.x already)_
- [ ] integration test on kernel
- [ ] jira issue

pspanja:
I'm taking this over, thanks for the contribution @florianalexandre!
Continued in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/80, the commit is preserved there.

PR #70, opened 2016-10-21
Fix EZP-26484: DocumentId/LocationId in index causes bad deletion beh…

siegfried-zach:
Issue: https://jira.ez.no/browse/EZP-26484
Adding delimiter between ID and languageCode, thus the delete queries are always unique to one documentId

### DIFF
--- a/lib/DocumentMapper/NativeDocumentMapper.php
+++ b/lib/DocumentMapper/NativeDocumentMapper.php
@@ -395,6 +395,12 @@ class NativeDocumentMapper implements DocumentMapper
      * The above is useful when targeting all Content's documents, without
      * the knowledge of it's translations.
      *
+     * In https://jira.ez.no/browse/EZP-26484 a bug was detected:
+     * At object deletions, this happens without language code, resulting in a delete query like:
+     * content1234* --> which deletes all IDs with bigger IDs like 12341, 12342, causing random deletion of objects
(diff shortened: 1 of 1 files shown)


lib/DocumentMapper/NativeDocumentMapper.php
```diff
@@ -395,14 +395,20 @@ public function mapContentBlock(Content $content)
      * The above is useful when targeting all Content's documents, without
      * the knowledge of it's translations.
      *
+     * In https://jira.ez.no/browse/EZP-26484 a bug was detected:
+     * At object deletions, this happens without language code, resulting in a delete query like:
+     * content1234* --> which deletes all IDs with bigger IDs like 12341, 12342, causing random deletion of objects
+     * Introducing a delimiter between ID and language code, thus the delete query only deletes occurances of given ID.
+     * Underscore as delimiter did not work - stripped out somewhere so using "lang" - should not matter
```
andrerom: Could simplify this as and combine with block above into something like:
```
Format of id is "content<content-id>lang[<language>]".
If $language code is not provided, the method will return prefix of the IDs
of all Content's documents (there will be one document per translation).
The above is useful when targeting all Content's documents, without
the knowledge of it's translations, and thanks to "lang" string it will not
risk matching other documents (as was the case in EZP-26484).
```

Similar for `generateLocationDocumentId`.

PR #71, opened 2016-11-24
EZP-26668: Index fields though field mapper plugins

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-26668

Following on https://jira.ez.no/browse/EZP-26368, this refactors `NativeDocumentMapper` implementation to use only field mappers to map document fields.

Duplication of internal methods in `ContentFieldMapper/ContentDocumentBaseFields` and `LocationFieldMapper/LocationDocumentContentFields` is intentional and temporary, since the followup to this is normalization of field names between Content and Location documents, which will enable replacing these with a common block document level field mapper.

### DIFF
--- a/lib/FieldMapper/ContentFieldMapper/ContentDocumentBaseFields.php
+++ b/lib/FieldMapper/ContentFieldMapper/ContentDocumentBaseFields.php
@@ -1,0 +1,246 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
+ */
+namespace EzSystems\EzPlatformSolrSearchEngine\FieldMapper\ContentFieldMapper;
+
+use EzSystems\EzPlatformSolrSearchEngine\FieldMapper\ContentFieldMapper as BaseContentFieldMapper;
+use EzSystems\EzPlatformSolrSearchEngine\DocumentMapper;
+use eZ\Publish\SPI\Persistence\Content\Location\Handler as LocationHandler;
+use eZ\Publish\SPI\Persistence\Content\Type\Handler as ContentTypeHandler;
+use eZ\Publish\SPI\Persistence\Content\ObjectState\Handler as ObjectStateHandler;
+use eZ\Publish\SPI\Persistence\Content\Section\Handler as SectionHandler;
+use eZ\Publish\SPI\Persistence\Content;
+use eZ\Publish\SPI\Search\Field;
+use eZ\Publish\SPI\Search\FieldType;
+
+/**
+ * Maps base Content related fields to a Content document.
+ */
+class ContentDocumentBaseFields extends BaseContentFieldMapper
+{
+    /**
+     * @var \eZ\Publish\SPI\Persistence\Content\Location\Handler
+     */
+    protected $locationHandler;
+
+    /**
+     * @var \eZ\Publish\SPI\Persistence\Content\Type\Handler
+     */
+    protected $contentTypeHandler;
+
+    /**
+     * @var \eZ\Publish\SPI\Persistence\Content\ObjectState\Handler
+     */
+    protected $objectStateHandler;
+
+    /**
+     * @var \eZ\Publish\SPI\Persistence\Content\Section\Handler
+     */
+    protected $sectionHandler;
+
+    /**
+     * @param \eZ\Publish\SPI\Persistence\Content\Location\Handler $locationHandler
+     * @param \eZ\Publish\SPI\Persistence\Content\Type\Handler $contentTypeHandler
+     * @param \eZ\Publish\SPI\Persistence\Content\ObjectState\Handler $objectStateHandler
+     * @param \eZ\Publish\SPI\Persistence\Content\Section\Handler $sectionHandler
+     */
+    public function __construct(
+        LocationHandler $locationHandler,
+        ContentTypeHandler $contentTypeHandler,
+        ObjectStateHandler $objectStateHandler,
+        SectionHandler $sectionHandler
+    ) {
+        $this->locationHandler = $locationHandler;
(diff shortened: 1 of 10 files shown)


pspanja:
Ready for review, ping @andrerom @alongosz

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentFulltextFields.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+namespace EzSystems\EzPlatformSolrSearchEngine\FieldMapper\ContentTranslationFieldMapper;
+
+use eZ\Publish\SPI\Search\Field;
+use eZ\Publish\SPI\Search\FieldType;
+
+/**
+ * Maps Content fulltext fields to Content document.
+ */
+class ContentDocumentFulltextFields extends BlockDocumentsContentFields
```
andrerom: What is this class doing?

andrerom: _(which isn't already done by parent)_

andrerom: > not 

aha, missed that :)

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentFulltextFields.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    protected function appendField(array &$fields, Field $documentField)
```
andrerom: please add doc block here that describes that it is overloaded since we need to match full text fields, as opposed to matching everything but fulltext field like parent.

lib/FieldMapper/ContentFieldMapper/ContentDocumentBaseFields.php
```diff
@@ -0,0 +1,244 @@
...
+                $this->getObjectStateIds($contentInfo->id),
+                new FieldType\MultipleIdentifierField()
+            ),
+        ];
+    }
+
+    /**
+     * Returns an array of object state ids of a Content with given $contentId.
+     *
+     * @param int|string $contentId
```
alongosz: Nitpick: I'm not a fan of hinting multiple types for arguments. It seems this argument is only used with ObjectState Handler's Gateway, which [hints int](https://github.com/ezsystems/ezpublish-kernel/blob/v6.6.0/eZ/Publish/Core/Persistence/Legacy/Content/ObjectState/Gateway/DoctrineDatabase.php#L558) anyway.

lib/FieldMapper/ContentFieldMapper/ContentDocumentLocationFields.php
```diff
@@ -0,0 +1,134 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
```
alongosz: Nitpick#2: to be skipped ;)

lib/FieldMapper/LocationFieldMapper/LocationDocumentContentFields.php
```diff
@@ -0,0 +1,252 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
```
alongosz: Nitpick#3: to be skipped.

lib/FieldMapper/LocationFieldMapper/LocationDocumentContentFields.php
```diff
@@ -0,0 +1,252 @@
...
+                $this->getObjectStateIds($contentInfo->id),
+                new FieldType\MultipleIdentifierField()
+            ),
+        ];
+    }
+
+    /**
+     * Returns an array of object state ids of a Content with given $contentId.
+     *
+     * @param int|string $contentId
```
alongosz: As I've mentioned before I would prefer this hinting only `int`.

lib/FieldMapper/LocationFieldMapper/LocationDocumentContentFields.php
```diff
@@ -0,0 +1,252 @@
...
+        return $objectStateIds;
+    }
+
+    /**
+     * Returns Content ids of all ancestor Locations of all Locations
+     * of a Content with given $contentId.
+     *
+     * Used to determine user groups of a user with $contentId.
+     *
+     * @param int|string $contentId
```
alongosz: Same comment as above.

lib/FieldMapper/ContentFieldMapper/ContentDocumentBaseFields.php
pspanja: This is because next storage engine might use strings for ID's. We don't use gateway here as those are storage specific, but ObjectState Handler used here hints `mixed`. Some time ago we agreed to use `int|string` instead, just did not update everything.

lib/FieldMapper/ContentFieldMapper/ContentDocumentLocationFields.php
pspanja: Removed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/71/commits/42266f5c5f4b27764ce81753aba9e2586a20b96a

lib/FieldMapper/LocationFieldMapper/LocationDocumentContentFields.php
pspanja: Removed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/71/commits/42266f5c5f4b27764ce81753aba9e2586a20b96a

pspanja:
Thanks for the reviews, merging when tests finish.

PR #72, opened 2016-11-25
EZP-26687: Consolidate field mapping and query visiting implementation

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-26687

Following on https://jira.ez.no/browse/EZP-26668, field mapping and query visiting implementation contains some duplication in regard to fields that are common to both document types (Content and Location).

This refactors implementation so that unnecessary duplication is removed.


pspanja:
Ready for review, ping @alongosz @andrerom

.travis.yml
```diff
@@ -24,6 +24,7 @@ branches:
     only:
         - master
         - /^\d.\d+$/
+        - EZP-26668-use-field-mappers
```
andrerom: tmp

pspanja: Removed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/72/commits/43667eda066521a2d1e409b2e9feb41cad13bced

emodric:
Isn't this a major BC break? Namespaces are changed, service names are changed, bundles using them will need to go through hoops to support 1.1 and 1.2 version of the bundle...

andrerom:
I wouldn't call it major, there is no interface or api changes here. Could you give an example?

emodric:
As far as I'm concerned, everything that is not explicitly marked private, protected or internal in docs/config/phpdoc is part of the "API".

Is it documented somewhere that these services and classes should not be used directly?

If you wish for an example: https://github.com/netgen/NetgenEnhancedSelectionBundle/blob/master/Core/Search/Solr/Query/CriterionVisitor/EnhancedSelection.php

This visitor was implemented recently using existing visitors as a guide, since no other documentation exists. Meaning, it uses abstract Field visitor which now had it's namespace changed. I call that a major BC break.

andrerom:
fair point, @pspanja do you see any way we can maintain extensibility BC for visitors? @emodric anything else then visitors you had in mind here as problematic for extending the bundle? _(release is done, so can't redo the whole change here, it's to late for that)_

emodric:
@pspanja should be able to answer better than me, but if we compare it to old eZ Find, then criterion visitors, sort clause visitors, facet builders and field mappers would be the main extension points for the bundle I guess.

pspanja:
We could have kept the abstract class on the same namespace, marking it as deprecated and making it extend the abstract class in the new namespace.

Otherwise, we should probably take some time to mark everything we don't consider a supported extension point as internal or make it `final` and `private` where possible. :)

andrerom:
> We could have kept the abstract class on the same namespace, marking it as deprecated and making it extend the abstract class in the new namespace

ok, could you do a PR on that?

> Otherwise, we should probably take some time to mark everything we don't consider a supported extension point as internal or make it final and private where possible. :)

Well. We haven't said we follow Symfony's interpretation of SemVer on bc policy. What makes sense for a framework does not nicely fit so well on applications and bundles. Just because PHP lacks private namespaces, doesn't mean everything should be extended. TBD

emodric:
> We haven't said we follow Symfony's interpretation of SemVer on bc policy.

Well, you haven't said you don't :) Anyway, there is no Symfony's interpretation on SemVer and BC. SemVer rules explicitly state that minor versions should be bumped only if you add features that don't break BC. Breaking BC means new major version, no exceptions.

pspanja:
Here we go https://github.com/ezsystems/ezplatform-solr-search-engine/pull/85 :)

andrerom:
> SemVer rules explicitly state that minor versions should be bumped only if you add features that don't break BC

SemVer talks about `Public API` when saying that.
It states nothing about internal classes in random packages being considered as "Public API", however symfony does include these in their BC promise: http://symfony.com/doc/current/contributing/code/bc.html

So I agree on things you are _supposed_ to extend, but not on things you don't. The distinction will never be 100% clear, there is always be a grey area here that we need to discover as we progress. _(same was the case in legacy btw, sometimes people extended things we did'nt plan to be extended, and then we where forced to define it)_

andrerom:
@emodric updated comment above ;)

emodric:
> SemVer talks about Public API when saying that.

Yes, but it doesn't define what public API is. That is left for bundle/lib authors to decide, and the problem is that there was no clear definition here.

> I also disagree with assuming everything is Public API, cause that makes it super hard to improve our internals. Even bug fixes becomes much more difficult to handle.

Well, nobody said that maintaining a product used by tens of thousands of people is easy ;) But on serious note, you put the bundle in open source, and say: here, use it, extend it, you must be prepared that people will do exactly that. As mentioned before, it wasn't clearly defined what official extension point is and what is not.

> But the solution to this grey area is perhaps that we mark everything as internal and selectively mark what is supposed to be extended (as opposed to decorate which uses interface and hence is clearly public api).

Yeah, that's what Symfony started doing for v4.

andrerom:
> Yeah, that's what Symfony started doing for v4.

Ok, didn't know that. In v2 they uses `@api`, which I kind of prefer, as it is easy to forget to add `@internal` but not so easy to forget to add `@api`. WDYT?

Doctrine is more anal on this and uses ´final´, it avoids the grey area, but maybe rigid for our context here. At least I would be annoyed if I ever needed to extend their classes to do a workaround\* until contribution upstream goes true.

\* effectively a hack I would be aware of having to adjust if upstream changes, it's the risk you do when you need to change logic of other peoples code. imo. But as said in #85, in this case I think you touched on a missing reusable component for this use case, so does not apply to this case.

anyway.. :)

emodric:
> Ok, didn't know that. In v2 they uses @api, which I kind of prefer, as it is easy to forget to add @internal but not so easy to forget to add @api. WDYT?

This is just a start, but it's there: https://github.com/symfony/symfony/pull/20493

As for `@api`, I guess it's fine, but I don't like annotations personally (except maybe in tests) as they are easily overlooked by developers, I would be more in favor of clear docs on Confluence.

andrerom:
> , I would be more in favor of clear docs on Confluence

Thats easily a broken workflow, would either generate this code info based on the code.

But maybe we are talking about different things, I of course agree we need to add more docs and examples of how to extend things. however we also need to mark on the code level so it's clear for us working on the code itself and people contribution to it.

PR #74, opened 2016-11-30
EZP-26707: Search with Solr does not respect isSearchable setting of fields

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-26707

This skips indexing full text data if the field is marked as not searchable (`FieldDefinitions::$isSearchable` property).
*Regular* fields are still indexed, since `isSearchable` is checked when resolving query for `Field` criterion/sort clause (and we're also going for `isSearchable` meaning only full text searchable).

### TODOs
- [x] integration test on https://github.com/ezsystems/ezpublish-kernel side

### DIFF
--- a/lib/DocumentMapper/NativeDocumentMapper.php
+++ b/lib/DocumentMapper/NativeDocumentMapper.php
@@ -470,10 +470,10 @@ class NativeDocumentMapper implements DocumentMapper
                         $indexField->type
                     );
 
-                    if ($documentField->type instanceof FieldType\FullTextField) {
-                        $fieldSets[$field->languageCode]['fulltext'][] = $documentField;
-                    } else {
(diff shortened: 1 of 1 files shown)


pspanja:
Since this was tested manually, I'll write the tests on https://github.com/ezsystems/ezpublish-kernel for master only, to avoid merges all the way from 6.0.

pspanja:
This now fails since for PHP 5.5 and 5.6 ezsystems/ezpublish-kernel (6.4.x-dev 8ec1f8d) is installed.
On the other hand, for PHP 5.4 this uses new micro tag ezsystems/ezpublish-kernel (v6.0.1.4), which passes and validates the fix.

Still waiting for https://github.com/ezsystems/ezpublish-kernel/pull/1841, which has tests.

PR #75, opened 2016-11-30
[Composer] Bump PHP requirements to be in sync with kernel

andrerom:


andrerom:
Verified locally, as @travis-ci is "down" _(afternoon flood)_, I'll merge to make sure we have this corrected.

PR #81, opened 2016-12-06
EZP-26774: switch from q.alt to q parameter to enable use of spellchecker in future version

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-26774

Counterpart on `ezpublish-kernel`: https://github.com/ezsystems/ezpublish-kernel/pull/1859

This switches used Solr query parameter from `q.alt` to `q`.
This is internal thing to the Solr backend, will enable using suggestions/spellchecking on the generated query.

Note: suggestions/spellchecking in this context is about handling it on the Solr backend side, not in the Solr search engine.

Related PR: https://github.com/ezsystems/ezplatform-solr-search-engine/pull/57

Also exposed and fixed:
- `MatchNone` visitor was not handled as subquery, which could result in unwanted pure negative query when used with other criteria
- removed unnecessary usage of `MatchAll` criterion
- refactored logical criteria visitors to avoid generating unnecessary parentheses

### DIFF
--- a/lib/CoreFilter/NativeCoreFilter.php
+++ b/lib/CoreFilter/NativeCoreFilter.php
@@ -103,13 +103,16 @@ class NativeCoreFilter extends CoreFilter
             $languageSettings['useAlwaysAvailable'] === true
         );
 
-        $query->filter = new LogicalAnd(
-            array(
-                new CustomField(self::FIELD_DOCUMENT_TYPE, Operator::EQ, $documentTypeIdentifier),
-                $query->filter,
-                $this->getCoreCriterion($languages, $useAlwaysAvailable),
-            )
-        );
+        $criteria = [
+            new CustomField(self::FIELD_DOCUMENT_TYPE, Operator::EQ, $documentTypeIdentifier),
+            $this->getCoreCriterion($languages, $useAlwaysAvailable),
+        ];
+
+        if ($query->filter !== null) {
+            $criteria[] = $query->filter;
+        }
+
+        $query->filter = new LogicalAnd($criteria);
     }
 
     /**
--- a/lib/Handler.php
+++ b/lib/Handler.php
@@ -187,7 +187,6 @@ class Handler implements SearchHandlerInterface
     public function findLocations(LocationQuery $query, array $fieldFilters = array())
     {
         $query = clone $query;
-        $query->filter = $query->filter ?: new Criterion\MatchAll();
         $query->query = $query->query ?: new Criterion\MatchAll();
 
         $this->coreFilter->apply(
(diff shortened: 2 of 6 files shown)


pspanja:
Ready for review, ping @andrerom @alongosz

andrerom:
Any reason why you ended up staying with defType edismax ? And doesn't this mean fullText on query is not using lucene parser?

pspanja:
> Any reason why you ended up staying with defType edismax ?

AFAIR the main reason we initially switched to it was support for boost query, otherwise right now, for what we use, lucene parser would fit just as well: https://github.com/ezsystems/ezplatform-solr-search-engine/pull/11.

> And doesn't this mean fullText on query is not using lucene parser?

No, tested it and it still uses Lucene parser, both for Content and Location search. Location search visitor uses Block Join parser with local params, the (sub) query there would still default to the standard/Lucene parser even if we forced Edismax on the main query. It would need to be explicitly forced on the subquery.

> And wouldn't it make sense to backport this?

Yes, I guess so :) I will rebase against `1.0`.

pspanja:
Now targets `1.0`.

andrerom:
Sorry, errors in the full text question. What I meant to ask was: isn't this change moving full text from edismax to lucene? _(along with everything else)_

pspanja:
No, because we always used Lucene parser for full text. [`q.alt`](https://cwiki.apache.org/confluence/display/solr/The+DisMax+Query+Parser#TheDisMaxQueryParser-Theq.altParameter) parameter (inherited to Edismax from Dismax) uses Lucene parser by default. So this moves the query to the `q` parameter, but keeps the same parser.

Aside, we should eventually use Edismax for full text search, it's what it was made for.

andrerom:
> Aside, we should eventually use Edismax for full text search, it's what it was made for.

I agree _(but we have seperate thread for that)_, then remaining questions is then:
- with lucene parser picked inline like this as opposed to set by default, will we be able to opt in to edismax for fulltext critera in the middel of it? or do you plan something else?
- with lucene parser now used for query and filter query, is boosting essentially disabled atm? Thought we where already using it for things like picking languages.

pspanja:
> with lucene parser picked inline like this as opposed to set by default, will we be able to opt in to edismax for fulltext critera in the middle of it? or do you plan something else?

Yes, we can change the parser using [local parameters](https://cwiki.apache.org/confluence/display/solr/Local+Parameters+in+Queries) with `v` key, we already do it with Block Join parser in [Fulltext visitor for Location search](https://github.com/ezsystems/ezplatform-solr-search-engine/blob/master/lib/Query/Location/CriterionVisitor/FullText.php#L34). It will also work in the subquery there, something like:

``` 
{!child of='document_type_id:content' v='document_type_id:content AND {!edismax v="..."}'}
```

> with lucene parser now used for query and filter query, is boosting essentially disabled atm? Thought we where already using it for things like picking languages.

We only use boosting in full text search, per field, that works with Lucene parser. We don't use it for picking languages. AFAIR this was discussed as one possible approach to solve that, but we do it differently in `CoreFilter` service.

Boost functionality I mentioned we switched for is [`Boost Query`](https://cwiki.apache.org/confluence/display/solr/The+DisMax+Query+Parser#TheDisMaxQueryParser-Thebq(BoostQuery)Parameter) (separate parameter `bq`). Eventually we can model that in API somehow. It should enable simple stuff like additionally boosting recent documents, but maybe also penalization of fallback languages or complex scoring rules for some kind of business app needs.

pspanja:
Okay, here we go!

PR #83, opened 2016-12-13
EZP-26775: supported Solr backend v6

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-26775

`ezpublish-kernel` counterpart: https://github.com/ezsystems/ezpublish-kernel/pull/1864

This refactors Travis Solr initialization script and adds ability to test Solr `6.3.0`.
This is still standalone/multicore setup, no Solr Cloud testing is done. It would be possible to set that up, but the implementation requires some work before support for it would be sensible (we would be limited to 1 replica per shard).

From default full text field configuration `enablePositionIncrements` flag is removed as it's invalid with Solr v5+, and it was not used anyway since Solr v4.3.
See [Stop Filter documentation](https://cwiki.apache.org/confluence/display/solr/Filter+Descriptions#FilterDescriptions-StopFilter) for more details.

### TODOs

- [x] update `README.md`
- [x] PR on `ezpublish-kernel`: https://github.com/ezsystems/ezpublish-kernel/pull/1864

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,11 +67,9 @@ For Contributing to this Bundle, you should make sure to run both unit and integ
 
 4. Start Solr
 
-    *Note: In this case in seperate terminal for debug use*
-
     ```bash
-    cd solr-4.10.4/example
-    java -Djetty.port=8983 -jar start.jar
+    cd solr-6.3.0
+    bin/solr start
     ```
 
 5. Run integration tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,11 +17,17 @@ matrix:
         - php: 7.0
           env: TEST_CONFIG="phpunit.xml"
         - php: 5.6
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="4.10.4" CORES_SETUP="dedicated" SOLR_CONFS="lib/Resources/config/solr/schema.xml lib/Resources/config/solr/custom-fields-types.xml lib/Resources/config/solr/language-fieldtypes.xml" SOLR_CORES="core0 core1 core2 core3"
+          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="4.10.4" CORES_SETUP="dedicated"
         - php: 5.6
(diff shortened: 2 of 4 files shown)


pspanja:
Ping @andrerom @alongosz

MarioBlazek:
+1

README.md
```diff
@@ -67,17 +67,30 @@ For Contributing to this Bundle, you should make sure to run both unit and integ
 
 4. Start Solr
 
-    *Note: In this case in seperate terminal for debug use*
+    To start Solr server execute:
 
     ```bash
-    cd solr-4.10.4/example
-    java -Djetty.port=8983 -jar start.jar
+    ./bin/.travis/init_solr.sh
```
andrerom: script not meant for having to work on anyone's machine and command lines, so rather not :)

Side: folders starting with `.` is meant to be internal, in this case specific for travis, but it could be for ci in general.

andrerom:
From my side I we don't really need to keep on supporting Solr 4.10, but from the looks of the PR looks ok to do both until 2.x. However what are the cleanup / improvment benefits by going Solr 6.x only?  _(I assume we can start to take advantage of some features if we do)_

README.md
pspanja: Good point :)
Amended the commit to just update for Solr v6 in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/83/commits/3343d6cdd6012276214ec6a8616a1903b9e6b196

FWIW, here's what I use to start Solr by version: https://gist.github.com/pspanja/202462dab3a4934948174be7a9bb4376

pspanja:
> However what are the cleanup / improvment benefits by going Solr 6.x only? (I assume we can start to take advantage of some features if we do)

What started this is https://jira.ez.no/browse/EZP-26325, so it should get scoring when using block-join fixed, for us that means scoring with Location search. I will need to add some integration tests for that.

In the future we could benefit from more stuff, like [MoreLikeThis query parser](https://issues.apache.org/jira/browse/SOLR-6248), and there are of course improvements and fixes across the board.

PR #85, opened 2016-12-16
Add deprecated implementation for BC reasons

pspanja:
Adds deprecated implementation for the abstract `Field` criterion visitor on the previous namespace.

See https://github.com/ezsystems/ezplatform-solr-search-engine/pull/72

### DIFF
--- a/lib/Query/Common/CriterionVisitor/Field.php
+++ b/lib/Query/Common/CriterionVisitor/Field.php
@@ -18,7 +18,9 @@ use eZ\Publish\SPI\Search\Field as SearchField;
 use eZ\Publish\SPI\Search\FieldType;
 
 /**
- * Visits the Field criterion.
+ * Base class for Field criterion visitors.
+ *
+ * @api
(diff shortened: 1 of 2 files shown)


andrerom:
+1

andrerom:
@pspanja note that the issue here is still that people want to reuse the field logic, maybe we can consider extracting that as a trait, or even mark the class as @ api actually, so people can use it if they want. if so the deprecation notice should point to the new class.

pspanja:
OK, now updated deprecation message and marked extended class as \@api.

Ping @emodric

andrerom:
+1

emodric:
+1

pspanja:
OK, here we go :)

pspanja:
[1.2.1](https://github.com/ezsystems/ezplatform-solr-search-engine/releases/tag/v1.2.1) tagged.

MarioBlazek:
+1

PR #90, opened 2017-04-05
EZP-27227: Implement index-time boosting

pspanja:
### JIRA issue: https://jira.ez.no/browse/EZP-27227

This implements index-time boost functionality, similar to the [one existing in eZ Find](https://github.com/ezsystems/ezfind/blob/master/settings/ezfind.ini#L215).

One problem found when implementing this was usage of [Solr's `copyField` directive](https://cwiki.apache.org/confluence/display/solr/Copying+Fields) to index full-text data to the `text` field. It was not possible to keep this, because in order for a field to have boost factor it must have [norms](https://cwiki.apache.org/confluence/display/solr/Field+Type+Definitions+and+Properties) (attribute `omitNorms="false"` in the schema). Field that has norms also must be indexed, which voids the whole approach with having non-indexed `*_fulltext` fields that are only copied to the destination `text` field.

For that reason `copyField` approach is replaced with collecting all full-text data and indexing it directly in a [`TextField` type field](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/SPI/Search/FieldType/TextField.php). In future [`FullTextField` type field](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/SPI/Search/FieldType/FullTextField.php) could be removed in favour of explicit `getFullTextData()` method on the FieldType's [`Indexable` interface](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/SPI/FieldType/Indexable.php). For more details see commit https://github.com/ezsystems/ezplatform-solr-search-engine/commit/9b1e9702733fddc6a3cab46975218bff3d06be85 where field mappers for Content fields were refactored.

Boost factors are provided by the [`BoostFactorProvider`](https://github.com/netgen/ezplatform-solr-search-engine/blob/index-time-boost/lib/FieldMapper/BoostFactorProvider.php) service, which is configured with boost factor map by the Symfony's semantic configuration. It has only two methods to get the boost factor:

- `getContentFieldBoostFactor(ContentType $contentType, FieldDefinition $fieldDefinition)`
- `getContentMetaFieldBoostFactor(ContentType $contentType, $fieldName)`

Boost factor configuration is a part of the connection configuration for the search engine. It's probably best described by an example:

```yaml
ez_search_engine_solr:
    connections:
        default:
            boost_factors:
                content_type:
                    article: 2.0
                field_definition:
                    title: 3.0
                    blog_post:
                        title: 1.5
                meta_field:
                    name: 10.0
                    article:
                        text: 5.0
```

Boost can be mapped for regular Content fields and for Content's meta fields. Two meta fields are recognized, `name` (translated Content name) and `text` (catch-all language analyzed field for full-text search). Names `name` and `text` are configuration-level names, mapped to the Solr backend fields which are not directly exposed. Boost factors can be mapped by the ContentType, FieldDefinition and meta field name.

With the configuration example above, the behavior is as follows:

- `getContentFieldBoostFactor('article', 'title')` returns `2.0`
- `getContentFieldBoostFactor('news', 'title')` returns `3.0`
- `getContentFieldBoostFactor('blog_post', 'title')` returns `1.5`
- `getContentFieldBoostFactor('news', 'description')` returns `1.0` (default)
- `getContentMetaFieldBoostFactor('article', 'text')` returns `5.0`
- `getContentMetaFieldBoostFactor('blog_post', 'name')` returns `10.0`
- `getContentMetaFieldBoostFactor('article', 'name')` returns `2.0`

That means mapping under `content_type` key acts both as a default and an override for the factor mapping under `field_definition` and `meta_field` keys. It acts as an override when the factors are mapped through the FieldDefinition identifier/meta field name only. On the other hand, it will be overriden by the mapping when the factor is mapped through the ContentType identifier as well.

We discussed this a bit in the office, and the agreement was that behavior is sensible, as mapping though a ContentType is more specific than through a FieldDefinition/field name.

### UpdateSerializer

`UpdateSerializer` is a service creating XML that is posted to the Solr's `/update` resource and it was extracted out of the Gateway implementation.

### Fulltext catch-all field name change

Name of the language analyzed field where all Content's text is indexed has been changed from `text` to `meta_content__all_text_t`. I did this because keeping the name would require special handling when creating the update XML. In retrospect it seems like it might be worth to keep the name, as it will mean a BC break if it was used directly. (Feedback welcome)

### Indexing translated Content name

A [dedicated field mapper](https://github.com/netgen/ezplatform-solr-search-engine/blob/index-time-boost/lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentTranslatedContentNameField.php) is implemented for indexing translated Content name. It will index the name both as a string (keyword) and text field (language analyzed and boosted).

Same as `text` meta field, `name` is indexed on the Content document only.

### Limitations

In order for boost factor to be applicable on the Solr field, the field must store norms (attribute `omitNorms="false"` in the schema). Otherwise Solr will respond with an exception, as was introduced in [SOLR-3241](https://issues.apache.org/jira/browse/SOLR-3241). This setting affects both the size of the storage and memory usage. In the schema norms are explicitly allowed only on the `text` type Solr field (language analyzed field). configured boosts will also be applied only on the field of that type through checking for [`TextField` type field](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/SPI/Search/FieldType/TextField.php).

In order for boosts to be used on other fields, custom field mappers will have to be implemented and schema will have to be modified so that norms can be stored on them. The goal of this is to provide a way to easily improve relevancy for full-text search by boosting on Content name and per ContentType, which should be enough for most cases. We can also improve this implementation when/if the needs become clear, but anything more than that, we would be much better off with a proper boost API using boost query functionality in `eDisMax` parser, than trying to force it through something that is quite limited by it's nature.

Also, index-time boosting needs to be limited to Solr, as [Elasticsearch does not support it anymore](https://www.elastic.co/guide/en/elasticsearch/reference/5.2/index-boost.html). So we should also go for [removing it from SPI](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/SPI/Search/FieldType.php#L46).

### Missing pieces

Currently the only way to target text-type field in Solr is through `CustomFieldCrierion`. Using that requires hardcoding field names that are indexed out of the box. Also, Content meta fields `text` and `name` are indexed on the Content document, so with those fields, it can work with Content search only. We should improve this somehow in a followup, some ideas:

- introduce `ContentName` criterion
- introduce `TEXT` operator on `Field` criterion (and a way to use different indexed field per operator)
- implement general purpose `Text` criterion that would work on Location search as well (through block join)
- provide a way to get meta field name and indexed FieldType field name in `FieldNameResolver`

### TODOs

- [x] ~~PR on https://github.com/ezsystems/ezpublish-kernel~~
  See https://github.com/ezsystems/ezpublish-kernel/pull/1948

### DIFF
--- a/bundle/ApiLoader/BoostFactorProviderFactory.php
+++ b/bundle/ApiLoader/BoostFactorProviderFactory.php
@@ -1,0 +1,66 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
+ */
+namespace EzSystems\EzPlatformSolrSearchEngineBundle\ApiLoader;
+
+use eZ\Bundle\EzPublishCoreBundle\ApiLoader\RepositoryConfigurationProvider;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+
+class BoostFactorProviderFactory implements ContainerAwareInterface
+{
+    use ContainerAwareTrait;
+
+    /**
+     * @var \eZ\Bundle\EzPublishCoreBundle\ApiLoader\RepositoryConfigurationProvider
+     */
+    private $repositoryConfigurationProvider;
+
+    /**
+     * @var string
+     */
+    private $defaultConnection;
+
+    /**
+     * @var string
+     */
+    private $boostFactorProviderClass;
+
+    /**
+     * @param \eZ\Bundle\EzPublishCoreBundle\ApiLoader\RepositoryConfigurationProvider $repositoryConfigurationProvider
+     * @param string $defaultConnection
+     * @param string $boostFactorProviderClass
+     */
+    public function __construct(
+        RepositoryConfigurationProvider $repositoryConfigurationProvider,
+        $defaultConnection,
+        $boostFactorProviderClass
+    ) {
+        $this->repositoryConfigurationProvider = $repositoryConfigurationProvider;
+        $this->defaultConnection = $defaultConnection;
+        $this->boostFactorProviderClass = $boostFactorProviderClass;
+    }
+
+    public function buildService()
+    {
+        $repositoryConfig = $this->repositoryConfigurationProvider->getRepositoryConfig();
+
+        $connection = $this->defaultConnection;
+        if (isset($repositoryConfig['search']['connection'])) {
+            $connection = $repositoryConfig['search']['connection'];
+        }
+
+        return new $this->boostFactorProviderClass(
+            $this->container->getParameter(
+                "ez_search_engine_solr.connection.{$connection}.boost_factor_map_id"
+            )
+        );
+    }
+}
--- a/bundle/ApiLoader/BoostFactorProviderFactory.php
+++ b/bundle/ApiLoader/BoostFactorProviderFactory.php
@@ -12,6 +12,9 @@ use eZ\Bundle\EzPublishCoreBundle\ApiLoader\RepositoryConfigurationProvider;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 
+/**
+ * BoostFactorProvider service factory takes into account boost factor semantic configuration.
+ */
 class BoostFactorProviderFactory implements ContainerAwareInterface
 {
     use ContainerAwareTrait;
--- a/bundle/ApiLoader/BoostFactorProviderFactory.php
+++ b/bundle/ApiLoader/BoostFactorProviderFactory.php
@@ -5,8 +5,6 @@
  *
  * @copyright Copyright (C) eZ Systems AS. All rights reserved.
  * @license For full copyright and license information view LICENSE file distributed with this source code.
- *
- * @version //autogentag//
  */
 namespace EzSystems\EzPlatformSolrSearchEngineBundle\ApiLoader;
 
--- a/bundle/DependencyInjection/Configuration.php
+++ b/bundle/DependencyInjection/Configuration.php
@@ -30,6 +30,11 @@ class Configuration implements ConfigurationInterface
         'path' => '/solr',
     );
 
+    protected $metaFieldNames = [
+        'name',
+        'text',
+    ];
+
     public function __construct($rootNodeName)
     {
         $this->rootNodeName = $rootNodeName;
@@ -93,6 +98,8 @@ class Configuration implements ConfigurationInterface
 
     /**
      * Adds connections definition.
+     *
+     * @throws \RuntimeException
      *
      * @param \Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition $node
      */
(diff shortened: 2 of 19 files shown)


MarioBlazek:
Wow, this one rocks, nice work @pspanja

MarioBlazek:
+1

hknezevic:
+1

mlukac:
+1

andrerom:
_Side: @mlukac @hknezevic @MarioBlazek please use the github reactions on the PR when you want to do a content less thumbs up / +1, lest keep the comments for discussions ;)_

Review ping @alongosz @Nattfarinn _(and bonus if available: @bdunogier @paulborgermans)_

pspanja:
Updated description clarifying that meta fields are indexed on the Content document only, so they can be targeted with `CustomFieldCriterion` in Content search only.

alongosz:
@pspanja small question:
> Also, index-time boosting needs to be limited to Solr, as [Elasticsearch does not support it anymore](https://www.elastic.co/guide/en/elasticsearch/reference/5.2/index-boost.html). So we should also go for [removing it from SPI](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/SPI/Search/FieldType.php#L46).

They actually say that:
> The index time boost mapping has been replaced with query time boost

So, instead of removing it from SPI, can't we just say that `$boost = 1` for Solr means index time boost and for ES means query time boost?

paulborgermans:
A (too small) comment: index time boosting is in general not the recommended way forward as it is only applicable when both your data domain model is rigid AND the queries represent an almost invariant view on the data. In any advanced use case (where you want to apply boosting of certain fields), this is not the case. Also, an "accurate" tuning of index time boost factors can only be done after a well executed round of query time boosting experiments.

Go for query time boosting only IMHO, for most eZ Publish implementations there is little performance overhead anyway

andrerom:
> Go for query time boosting only IMHO

@paulborgermans I think I lean towards agreeing on that, especially given Elastic has dropped support for index time boosting and we aim to try to find common middle ground when possible.

@pspanja any thoughts on how we can combine configuration based boosting and dynamic boosting? And furthermore would it then  make sense to move the config to kernel so it's applied to query by search service?

ilukac:
How many eZ projects use Elastic Search currently?  
When will eZ support Elastic Search, realistically?
How to make search results in eZ projects meaningful out of the box, now?
What is the practical difference for a developer between boosting configuration for index time and a boosting configuration for query time?

Lets be practical.

pspanja:
Hi guys, thanks for commenting 😄 

I acknowledge all that you say @paulborgermans. Index-time boost has not been recommended for some time now. But the problem is we don't support query-time boosting yet, and implementing a proper boost API seems like a bit bigger task. I also find that index-time boosting does fit nicely with the kind of segmented content we normally have with eZ Platform. While limited, it solves a number of use-cases for us, so for me it does have it's place, at least until we provide something better.

Regarding query-time boost, I think we need an API and extension points around `bq` (Boost Query) and `bf` (Boost Functions) parameters of eDisMax (and similar for Elasticsearch, should be investigated). Whether it would make sense to connect that with boosting configuration, currently I don't think so, exactly because it implies modifying the query :)

bundle/ApiLoader/BoostFactorProviderFactory.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
```
alongosz: nitpick, still occurs in some files, but maybe we should fix all of these at once?

pspanja:
@alongosz re Solr exception, schema needs to be updated, see https://github.com/ezsystems/ezplatform-solr-search-engine/pull/90/files#diff-2ef3da3ce7e2d13c93a0e1a4742c06c5

andrerom:
> How many eZ projects use Elastic Search currently?
> When will eZ support Elastic Search, realistically?

I hope none, undefined. Elastic is irrelevant for you for project perspectives so you can safely ignore that point. I was talking from architecture perspective, which is long term perspective.

> What is the practical difference for a developer between boosting configuration for index time and a boosting configuration for query time?

From the [Elasitc search doc](https://www.elastic.co/guide/en/elasticsearch/reference/5.2/mapping-boost.html) @pspanja linked to, but which should for the most part also apply to Solrm, doing it on query time will avoid the following:
- You cannot change index-time boost values without reindexing all of your documents.
- Every query supports query-time boosting which achieves the same effect. The difference is that you can tweak the boost value without having to reindex.
- Index-time boosts are stored as part of the norm, which is only one byte. This reduces the resolution of the field length normalization factor which can lead to lower quality relevance calculations.


We have almost another 2 months before next feature freeze, so it should be more then enough time to do adjustment of rather using query time boost internally in Solr Bundle, and leaving the rest of the logic here intact including config.

@pspanja: any idea if such a change would be big or not?

ilukac:
I am aware of query time boosting advantages. But there are advantages of index-time-boosting as well. The difficult part of this work was not the boosting part it self, rather the work around indexing fields separately (not in one single solr field) which is required for any kind of boosting. 

We closed this feature internally as it work as expected for several projects already. Actually it was very important for any kind of meaningful search results. 

IMO this PR should be good basis for adding more boosting features including query time but I don't think we have the time/budget to work on it now. I'll discussed this with @pspanja and its not a day or two of work :/

pspanja:
Hi @andrerom, sorry for taking so long with replying.

> We have almost another 2 months before next feature freeze, so it should be more then enough time to do afaik small adjustment of rather using query time boost internally in Solr Bundle, and leaving the rest of the logic here intact including config.
> It would be internal, so no need to wait for boosting to be better defined from API side for instance.
> any idea if such a change would be big or not?

To replicate what is done here by index-time boosting would require applying query-time boost on a meta field and on a regular Content text-type field, per ContentType. I think the mechanics of that would have to be quite awkward. A better option would be having a `Text` criterion with boost param + separate boosting on a field.

It would work a bit differently, but could replace index-time boost. But I would not go for mapping the configuration to the query-time boost, if the API is there, it should rather be used directly.

As @ilukac already said, we don't have a budget to go for that now. I'll probably go for some of it in my own spare time, but that tends tends to stretch out a bit (see https://github.com/netgen/query-translator for example).

bundle/DependencyInjection/Configuration.php
```diff
@@ -241,6 +248,112 @@ function ($v) {
                             ->end()
                         ->end()
                     ->end()
+                    ->arrayNode('boost_factors')
+                        ->addDefaultsIfNotSet()
+                        ->info('Index-time field boost factor mapping.')
```
andrerom: Please expand to something like.
```
Index-time field boost factor mapping.

Note: Changes to this configuration is not reflected on solr index without manually re-indexing the affected content or executing a full re-index. A future version might change this to apply on query time instead to avoid this.
```

lib/FieldMapper/BoostFactorProvider.php
```diff
@@ -0,0 +1,149 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
```
andrerom: Please remove these

andrerom:
So assuming you don't have the time to change it to be done at query time, it's not something we can prioritize to spend time on atm either. So as a compromise we should better document the pitfalls with index time boost for now and hint about it receiving changes in the future.

@ilukac / @pspanja  ok with that? :)

pspanja:
Fine with me :) I'll try to take care of review remarks today.

lib/FieldMapper/BoostFactorProvider.php
pspanja: Removed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/90/commits/19c368f1b2c581ee053b7dc41452ba33ec1567f2.

bundle/ApiLoader/BoostFactorProviderFactory.php
pspanja: Removed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/90/commits/19c368f1b2c581ee053b7dc41452ba33ec1567f2. I'll clean up the rest in a separate PR.

bundle/DependencyInjection/Configuration.php
pspanja: Expanded in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/90/commits/f3d86e2f4b6c0923a79add56b8ce5c69b7e9b52b.

bundle/ApiLoader/BoostFactorProviderFactory.php
```diff
@@ -0,0 +1,66 @@
...
+ */
+namespace EzSystems\EzPlatformSolrSearchEngineBundle\ApiLoader;
+
+use eZ\Bundle\EzPublishCoreBundle\ApiLoader\RepositoryConfigurationProvider;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+
+class BoostFactorProviderFactory implements ContainerAwareInterface
```
pspanja: Added in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/90/commits/bb81ebad2a06950c0457b7900eec39f5202c912c.

pspanja:
Resolved all remarks and rebased on top of `master`.

pspanja:
Reverted temporary commit in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/90/commits/b67698968c01a4cf28b1532e8a77be0b5cc14aa0. All done now :)

andrerom:
> Reverted temporary commit in ezsystems/ezplatform-solr-search-engine@b676989. All done now :)

@pspanja kernel change has been merged, so you can adapt composer like this:
`"ezsystems/ezpublish-kernel": "~6.7.4@dev|^6.9.1@dev|^7.0@dev"`

And then if all is correct tests will pass again :)

PR #92, opened 2017-05-09
EZP-27327: Fix User / ContentType / Section facet support in Solr

andrerom:
> Issue: https://jira.ez.no/browse/EZP-27327
> Epic: https://jira.ez.no/browse/EZP-26465
> Alternative to: #91*

Done:
- New logic for facet builder visitors with BC for existing once to be able to identify* which facets belongs to which facet builder, to be able to inject right facet builder when extracting the result from Solr:
  - This makes ContentType and Section facets fully implemented _(for now)_ as we can now use `name` from the user provided value in FacetBuilder.
-  Improve User Facet handling to support all three sub types: `modifier`, `owner` and `group`
- Don't throw when encountering unsupported FacetBuilders, just ignore, user will have to iterate returned facets to generate his UI/app
- Enable the 3 facets _(User, ContentType, Section)_ supported atm also on Location search

\* _The benefit of this approach over the one in #91 is that it  reduces complexity of facet builder visitors by getting them to set id on the facet, which in turn better gurantees we use right facet builder with a given facet during extraction phase, this was proposed by @kore in https://github.com/ezsystems/ezplatform-solr-search-engine/commit/274875f6379da52ea0f8515b6de6412503b924cb#commitcomment-22050214_

### DIFF
--- a/lib/ResultExtractor.php
+++ b/lib/ResultExtractor.php
@@ -10,7 +10,7 @@
  */
 namespace EzSystems\EzPlatformSolrSearchEngine;
 
-use EzSystems\EzPlatformSolrSearchEngine\Query\FacetBuilderVisitor;
+use EzSystems\EzPlatformSolrSearchEngine\Query\FacetFieldVisitor;
 use eZ\Publish\API\Repository\Values\Content\Search\SearchResult;
 use eZ\Publish\API\Repository\Values\Content\Search\SearchHit;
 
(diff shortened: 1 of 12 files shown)


lib/ResultExtractor.php
```diff
@@ -39,7 +39,7 @@ public function __construct(FacetBuilderVisitor $facetBuilderVisitor)
      *
      * @return \eZ\Publish\API\Repository\Values\Content\Search\SearchResult
      */
-    public function extract($data)
+    public function extract($data, array $facetBuilders = [])
```
andrerom: fixed in 2d6496d

PR #95, opened 2017-07-03
EZP-26806: Improve fulltext search query syntax

pspanja:
This resolves https://jira.ez.no/browse/EZP-26806
Related PR: https://github.com/ezsystems/ezpublish-kernel/pull/2101

This implements improved query syntax for FullText criterion.
Quick cheat sheet:

`word` `"phrase"` `(group)` `+mandatory` `-prohibited` `AND` `&&` `OR` `||` `NOT` `!`

Examples:

- grouping
   ```
   one AND (two AND (three OR four))
   ```

- phrases
   ```
   "this is a phrase" and these are words
   ```

- mandatory/prohibited operators (`+`, `-`)
   ```
   +one two -three
   ```

And so on.

This is achieved through https://github.com/netgen/query-translator library, using `text` variant of the syntax it provides. The syntax can be customized by adjusting regular expressions used to tokenize the query input, meaning it can be easily adjusted if needed (for example using only the subset of the syntax, changing the operators and so on). The produced query now targets `edismax` query parser.

I would suggest trying the demo of https://github.com/netgen/query-translator to better understand what this is.

### Todos
- [x] API Fulltext documentation
  - Probably need to define a simple and advance format in the doc, and a way for app to capability detect if service engine support advance format _(via search service for instance, pulling that from engine)_
- [x] ~~Elasticsearch implementation~~
- [x] Integration tests

### DIFF
--- a/lib/Query/Common/QueryTranslator/Generator/WordVisitor.php
+++ b/lib/Query/Common/QueryTranslator/Generator/WordVisitor.php
@@ -8,7 +8,7 @@
  *
  * @version //autogentag//
  */
-namespace EzSystems\EzPlatformSolrSearchEngine\Query\Common\QueryTranslator;
+namespace EzSystems\EzPlatformSolrSearchEngine\Query\Common\QueryTranslator\Generator;
 
 use QueryTranslator\Languages\Galach\Generators\Common\Visitor;
 use QueryTranslator\Languages\Galach\Generators\Lucene\Common\WordBase;
--- a/lib/Query/Common/QueryTranslator/Generator/WordVisitor.php
+++ b/lib/Query/Common/QueryTranslator/Generator/WordVisitor.php
@@ -5,8 +5,6 @@
  *
  * @copyright Copyright (C) eZ Systems AS. All rights reserved.
  * @license For full copyright and license information view LICENSE file distributed with this source code.
- *
- * @version //autogentag//
  */
 namespace EzSystems\EzPlatformSolrSearchEngine\Query\Common\QueryTranslator\Generator;
 
--- a/lib/Resources/config/container/solr/query_translator.yml
+++ b/lib/Resources/config/container/solr/query_translator.yml
@@ -37,7 +37,7 @@ services:
         class: QueryTranslator\Languages\Galach\Generators\Lucene\Common\Query
 
     ezpublish.search.solr.query.query_translator.galach.generator.visitor.word:
-        class: EzSystems\EzPlatformSolrSearchEngine\Query\Common\QueryTranslator\WordVisitor
+        class: EzSystems\EzPlatformSolrSearchEngine\Query\Common\QueryTranslator\Generator\WordVisitor
 
     ezpublish.search.solr.query.query_translator.galach.generator.common.aggregate:
         class: QueryTranslator\Languages\Galach\Generators\Common\Aggregate
--- a/lib/Resources/config/container/solr/query_translator.yml
+++ b/lib/Resources/config/container/solr/query_translator.yml
@@ -7,7 +7,7 @@ services:
     ezpublish.search.solr.query.query_translator.galach.tokenizer:
         class: QueryTranslator\Languages\Galach\Tokenizer
         arguments:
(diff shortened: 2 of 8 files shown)


pspanja:
If this is acceptable I will also implement it for Elasticsearch and add API doc and integration tests.

andrerom:
This looks promising, I'll try to look for the issue we created for this (I think) tomorrow, if you happen to know which one update the PR ;)

pspanja:
Found it: https://jira.ez.no/browse/EZP-26806
It actually describes having different syntax per engine. That was before I actually tried it and found that controlling all the features offered would still be ridiculous. The idea with this is to have common syntax, so the issue should be updated.

andrerom:
@pspanja ok, will you fix unit tests so we can start review? ;)

Also on the `netgen/query-translator` lib, did you evaluate others before deciding to write something new for it? :)

pspanja:
@andrerom Of course :) I wrote my own because those I found were either doing too little or too much, were not well tested, not maintained, not easy to understand and customize or not producing AST.

Re Legacy engine, it's of course possible to write SQL generator, but not sure if it would make sense... maybe some kind of simplified syntax could work there, I'll need to look into it.

I'm back on this in the evening, on other things now.

pspanja:
Unit tests are now updated.

lib/Query/Common/QueryTranslator/Generator/WordVisitor.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
```
pspanja: Removed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/95/commits/9e4645784e85b03ca1680c94abcc381ad64edfb5

lib/Resources/config/container/solr/query_translator.yml
```diff
@@ -0,0 +1,59 @@
...
+                - '@ezpublish.search.solr.query.query_translator.galach.generator.visitor.mandatory'
+                - '@ezpublish.search.solr.query.query_translator.galach.generator.visitor.phrase'
+                - '@ezpublish.search.solr.query.query_translator.galach.generator.visitor.prohibited'
+                - '@ezpublish.search.solr.query.query_translator.galach.generator.visitor.query'
+                - '@ezpublish.search.solr.query.query_translator.galach.generator.visitor.word'
+
+    ezpublish.search.solr.query.query_translator.galach.generator.edismax:
+        class: QueryTranslator\Languages\Galach\Generators\ExtendedDisMax
+        arguments:
+            - '@ezpublish.search.solr.query.query_translator.galach.generator.common.aggregate'
```
pspanja: Replaced with double quotes in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/95/commits/d9d34503eebe91ff4da192952e639c22787340a1

pspanja:
Thanks for the review so far, I'll proceed with todos.
Have in mind it might take me few days :)

andrerom:
@pspanja Any of the todo's we can descope here to get moving on this? like for instance elastic support can be done as followup given the aim to upgrade it's supported version anyway.

pspanja:
@andrerom It's mostly about the tests missing now, ES should not be a problem but we can skip it if it turns out to be a one anyway. I'm now back from vacation so will try to do this through this week or weekend at latest.

andrerom:
Ok don't hesitate to send pings, this is a really good change to get out, so if you get time this week we can aim for 6.12-beta otherwise would need to wait for 6.13.

andrerom:
I'll handle the capability part separately as it will go across versions, and @DominikaK take some time with @alongosz so he can show you what this enables and ping for input on doc on this :)

pspanja:
@andrerom thanks 😄

PR #96, opened 2017-07-28
Fix-EZP-27724: Unable to run dockerized Solr

kmadejski:
JIRA issue: [https://jira.ez.no/browse/EZP-27724](https://jira.ez.no/browse/EZP-27724)

# Description
As described in issue, run Solr is impossible because in the newest version of Solr ``6.6.0`` field ``defaultSearchField`` is deprecated. 
This PR removes mentioned field from ``schema.xml``.

Second solution proposed here: https://github.com/ezsystems/ezplatform/pull/196

### DIFF
--- a/lib/Resources/config/solr/schema.xml
+++ b/lib/Resources/config/solr/schema.xml
@@ -109,5 +109,4 @@ should not remove or drastically change the existing definitions.
     <field name="_version_" type="long" indexed="true" stored="true" multiValued="false" />
 
   <uniqueKey>id</uniqueKey>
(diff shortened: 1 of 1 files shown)


andrerom:
I prefer this one, as we aim to support 6.x. But is there any behaviour change here?

Also note that title can be much more to the point, it's not something we always remember, but when in general try to write something that makes sense for changelogs. Like `EZP-27724: Remove Solr 6.6 deprecated shema field stopping automated use` or something similar.

PR #98, opened 2017-08-01
EZP-27712: Enable RemoveTranslation Search Engine Slot

alongosz:
Implements [EZP-27712](https://jira.ez.no/browse/EZP-27712) for Solr Bundle
----

This PR adds `RemoveTranslation` Search Engine Slot which reindexes Content on `RemoveTranslationSignal` introduced in ezsystems/ezpublish-kernel#2031

*Note: for ES/SQL Slot see ezsystems/ezpublish-kernel#2060.*

**TODO**:
- [x] Implement and enable `RemoveTranslation` Slot for Solr
- [x] Ensure 1.7 EE LTS BC (see [comment](#issuecomment-319682941)).
- [x] Remove TMP commit before merge

### DIFF
--- a/tests/lib/Slot/RemoveTranslationTest.php
+++ b/tests/lib/Slot/RemoveTranslationTest.php
@@ -1,0 +1,90 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformSolrSearchEngine\Tests\Slot;
+
+use eZ\Publish\Core\SignalSlot\Signal\ContentService\RemoveTranslationSignal;
+use eZ\Publish\SPI\Persistence\Content\ContentInfo;
+use eZ\Publish\SPI\Persistence\Content as SPIContent;
+use EzSystems\EzPlatformSolrSearchEngine\Slot\RemoveTranslation;
+
+/**
+ * RemoveTranslation Slot Test.
+ */
+class RemoveTranslationTest extends TestCase
+{
+    /**
+     * @var \EzSystems\EzPlatformSolrSearchEngine\Slot\RemoveTranslation
+     */
+    private $slot;
+
+    /**
+     * Check if required signal exists due to BC.
+     */
+    public static function setUpBeforeClass()
+    {
+        if (!class_exists(RemoveTranslationSignal::class)) {
+            self::markTestSkipped('RemoveTranslationSignal does not exist');
+        }
+    }
+
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->slot = new RemoveTranslation(
+            $this->getRepositoryMock(),
+            $this->getPersistenceHandlerMock(),
+            $this->getSearchHandlerMock()
+        );
+    }
+
+    /**
+     * Test receiving RemoveTranslationSignal.
+     *
+     * @covers \EzSystems\EzPlatformSolrSearchEngine\Slot\RemoveTranslation::receive
+     */
+    public function testReceive()
+    {
+        $contentHandlerMock = $this->getContentHandlerMock();
+
+        $contentHandlerMock
+            ->expects($this->once())
+            ->method('loadContentInfo')
+            ->willReturn(
+                new ContentInfo(
+                    [
+                        'id' => 2,
+                        'currentVersionNo' => 1,
+                        'isPublished' => true,
+                    ]
+                )
+            )
+        ;
+
+        $content = new SPIContent();
+
+        $contentHandlerMock
+            ->expects($this->once())
+            ->method('load')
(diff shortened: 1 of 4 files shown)


andrerom:
I'm assuming this will work also on 6.7.x as it's purely an event (slot), and symfony won't complain the service here points to non existing class. right?

alongosz:
> I'm assuming this will work also on 6.7.x (given composer requriment allows that) as it's purely an event (slot), and symfony won't complain the service here points to non existing class. right?

This is actually what I plan to check tomorrow on LTS :-) 

Preliminary local PoC on clean Sf 2.8 showed no problems. It crashes on Sf 3.3 when service is defined using class name, but that's not the issue for LTS.
If we were to bump Solr Bundle we'd have to make it `2.0` as requirement on EE LTS is `^1.1` which matches anything `>= 1.1` and `< 2.0`. Change is to small to do it though TBH.

I'll know more tomorrow.

andrerom:
Ok, then it might be ok for normal installs then but let's see what your testing says. For those already using symfony 3 it might be an issue though.

Technically we could bump kernel requirement to 6.11, but that would imply yet another branch to maintain _(1.0, 1.4 & 1.latests)_. @emodric might have input on if this will affect them and if so alternative solutions.

emodric:
How about enabling it only if signal class exists, checking for it in the DI extension?

alongosz:
Unfortunately on LTS it crashes (class not found) when search service is used (even when signal is not emitted). Warm-up goes smoothly because, when building container, there's no check if class exists.

> How about enabling it only if signal class exists, checking for it in the DI extension?

I tried to come up with cleanest way to do it, but still don't like it very much:
Between [these lines](https://github.com/ezsystems/ezpublish-kernel/blob/v6.7.4/eZ/Publish/Core/Base/Container/Compiler/Search/SearchEngineSignalSlotPass.php#L45-L46) we'd have to place the following nasty ifs:
```php
// make sure class is defined (services created by factory don't need it) 
if (($class = $container->getDefinition($id)->getClass())) {
    // resolve parametrized class name
    $class = $parameterBag->resolveValue($class);
    if (!class_exists($class)) {
        continue;
    }
}
```

WDYT @andrerom @emodric ?

IMHO the best option is to keep `1.x` for `1.7 LTS` and create `2.0` for our latest stable, also requiring it in composer :(

andrerom:
> IMHO the best option is to keep 1.x for 1.7 LTS and create 2.0 for our latest stable, also requiring it in composer :(

I'm not a big Fan

The problem here is the usage of common slots, we will run into this every time we add a new signal.
Would rather go for this: The slot class is simple, lets create one here, and do one of two:
- depreciate it stating we will start using the one provided by kernel in 2.0
- intend that we stop using common slots altogether as it leads to unnecessary coupling which is intended to avoid by having an even system like signals slot in the first place.

andrerom:
I realize I forgot a third option: extracting the common search slots _(and maybe other stuff)_ into own package so it can be depended on independent of kernel versions _(this part would be important, otherwise same issue would appear again)_.

alongosz:
> B: intend that we stop using common slots altogether as it leads to unnecessary coupling, which an event system like signals slot tries to avoid in the first place.

In 2ea23c0 I implemented option B. Seems to be working for `1.7 LTS`. I successfully reindexed content, so everything that is supposed to be loaded loads correctly. `receive` is never executed, so won't complain on reference to non-existing `RemoveTranslationSignal`.

This kind of made sense because some Slots can be implemented differently - I avoided there loading locations as there are nested when indexing Content in case of Solr.

Things to resolve for 7.0:
- `eZ\Publish\Core\SignalSlot\Signal`
This should be API if we intend for package to have own Slots.
- `eZ\Publish\Core\Search\Common\Slot`
This should be SPI for the same reason.

> third option: extracting the common search slots (and maybe other common stuff) into own package so it can be depended on independent of kernel versions (this part would be important, otherwise same issue would appear again).

I like this idea also, however in this case loving option B more due to Locations being handled differently Solr. But still some common Slots seems like a valid direction.

BTW. I got bored tonight, so added unit tests for new Slot in 305d680. Could be easily removed if you think it's an overkill :D

alongosz:
Removed TMP commit after review. Won't matter for @Nattfarinn as he haven't seen this yet ;)

alongosz:
Rebased onto `1.4`, squashed "Slot" commits.

tests/lib/Slot/RemoveTranslationTest.php
```diff
@@ -0,0 +1,80 @@
...
+
+        $this
+            ->getSearchHandlerMock()
+            ->expects($this->once())
+            ->method('indexContent')
+            ->with($content)
+        ;
+
+        $this->slot->receive(
+            new RemoveTranslationSignal(['contentId' => 2, 'languageCode' => 'eng-US'])
```
andrerom: Same here I guess, we should skip if signal class does not exits.

alongosz: Silently amended in 8789d1c ;)

PR #100, opened 2017-08-14
EZP-27774: Basic HTTP Authorisation fix

sklimaszewski:
JIRA issue: https://jira.ez.no/browse/EZP-27774

## Description
As described in JIRA, the URL was built incorrectly (missing [at] sign) and `empty($this->username)` statement was wrong as well.

Futhermore, I have moved authorisation URL part `user:pass@` to the `getIdentifier()` method as this also has to be passed in `shards` URL parameter.


andrerom:
Thanks @sklimaszewski :)

PR #101, opened 2017-08-24
[WIP] EZP-27761: Translations are not being indexed/found by Solr

alongosz:
# [WiP] Fixes [EZP-27761](https://jira.ez.no/browse/EZP-27761)

This PR attempts to fix inconsistent behavior with respect to SQL search when no languageFilter settings were given when searching for Content.

While this solves the issue at hand, it introduces regression - Content defined in multiple languages is returned multiple times, so it needs to be investigated further.

**TODO**:
- [x] Match all languages (instead of main) when no languages were given
- [ ] Decide how to fix an issue with the same content returned multiple times.
- [ ] Remove TMP commit.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": "~5.6|~7.0",
-        "ezsystems/ezpublish-kernel": "~6.7.4@dev|^6.9.1@dev|^7.0@dev"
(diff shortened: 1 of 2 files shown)


andrerom:
@alongosz maybe you and @pspanja should talk about this and the Fulltext work this week ;)

alongosz:
Closing in favor of #168

PR #102, opened 2017-09-13
Enables PHPUnit v5

MarioBlazek:
This PR resolves:
* upgrades PHPUnit to version 5.7
* removes deprecated  `getMock()` method
* upgrades `matthiasnoback/symfony-dependency-injection-test` to version 1.0
* adds bundle and lib test to `autoload-dev` section of Composer autoload

PR #103, opened 2017-09-20
EZP-27916: [Integration tests] Make SetupFactory autoloaded in root packages

alongosz:
# Fixes [EZP-27916](https://jira.ez.no/browse/EZP-27916)

This PR fixes failures in CI tests for Solr run from `ezpublish-kernel` as a root-package.

Background story: The PR #102 splitted `autoload` settings into `autoload` and `autoload-dev`. 
This is fine when running tests from this package itself, however our `ezpublish-kernel` uses `SetupFactory` that exists in moved namespace prefix.

According to the [composer docs on `autoload-dev`](https://getcomposer.org/doc/04-schema.md#autoload-dev) this is `root-only` setting, so when Solr Bundle is required as a dependency, `composer dump-autoload` won't create proper paths.

**TODO**:
- [x] Discuss possible solutions.
- [x] Move `LegacySetupFactory` to `lib` and change its namespace.
- [x] Discuss the above approach
- [x] Simplify building of container inside `LegacySetupFactory` by reusing container build by the base `SetupFactory`
- [x] Create ezpublish-kernel PR counterpart with alignment of SetupFactory FQCN (ezsystems/ezpublish-kernel#2093)
- [x] Cleanup commits before merge

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
     "autoload": {
         "psr-4": {
             "EzSystems\\EzPlatformSolrSearchEngine\\": "lib",
-            "EzSystems\\EzPlatformSolrSearchEngineBundle\\": "bundle"
+            "EzSystems\\EzPlatformSolrSearchEngineBundle\\": "bundle",
+            "EzSystems\\EzPlatformSolrSearchEngine\\Tests\\SetupFactory\\": "tests/lib/SetupFactory"
         }
     },
     "autoload-dev": {
--- a/tests/lib/SetupFactory/LegacySetupFactory.php
+++ b/tests/lib/SetupFactory/LegacySetupFactory.php
@@ -11,12 +11,12 @@
 namespace EzSystems\EzPlatformSolrSearchEngine\Tests\SetupFactory;
 
 use eZ\Publish\API\Repository\Tests\SetupFactory\Legacy as CoreLegacySetupFactory;
-use eZ\Publish\Core\Base\ServiceContainer;
 use eZ\Publish\Core\Base\Container\Compiler as BaseCompiler;
 use EzSystems\EzPlatformSolrSearchEngine\Container\Compiler;
 use PDO;
 use RuntimeException;
 use eZ\Publish\API\Repository\Tests\SearchServiceTranslationLanguageFallbackTest;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
 use Symfony\Component\Config\FileLocator;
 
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -21,12 +21,12 @@
     "autoload": {
         "psr-4": {
             "EzSystems\\EzPlatformSolrSearchEngine\\": "lib",
-            "EzSystems\\EzPlatformSolrSearchEngineBundle\\": "bundle"
+            "EzSystems\\EzPlatformSolrSearchEngineBundle\\": "bundle",
+            "EzSystems\\EzPlatformSolrSearchEngine\\Tests\\": "tests/lib"
```
alongosz: I moved back relevant Test namespace to `autoload`. Due to cross-dependency hell ;) I don't have any other idea how to solve this (see PR description), but I'm open to suggestions.

// cc @MarioBlazek

andrerom: Which classes will be needed by kernel integration test suite? all or just a few?


Either way, make sure to be clear on why you make this change in the commit summary and description, someone will surely wonder why this is like this at some point and consider changing it back :P _at least if done like we do here_

MarioBlazek: @alongosz wouldn't actually be better to move `SetupFactory` to `lib` as it is meant to be reused by other libs ?

alongosz: > wouldn't actually be better to move SetupFactory to lib as it is meant to be reused by other libs?

Maybe :) My main issue with this approach is that SetupFactory is only a part of tests. I'll do it tomorrow and see how it looks.

alongosz:
Well... I moved SetupFactory outside of `lib` to the new namespace `\EzSystems\EzPlatformSolrSearchEngine\SetupFactory`

It required Kernel counterpart PR (see ezsystems/ezpublish-kernel#2093) backported to 6.7 LTS as Solr Bundle 1.4.x still supports it.
Solr failures are due to strict requirement for Kernel `6.7.x` on Travis (I probably can amend TMP commit to fix that if you want to see all green ;) )

WDYT about this solution @andrerom @MarioBlazek @pspanja ?

Side note:
Moreover, in 7c06763 I introduced a simplified way of building container for tests which reuses base SetupFactory from kernel. 
Reason: today (and not for the first time) I struggled a lot to make tests work locally with my weird symlinked setup of vendor repositories. This made me finally discover that it's enough just to override `SetupFactory\Legacy::externalBuildContainer` and dependency on relative path to Kernel's `config.php` magically disappears :)
It think this change is for the better.

andrerom:
Updating things across both repos are painful _(version conflicts, and unnecessary composer requirements bumps)_, why not just add something like the following to autoload array so kernel can stay the same across all versions:
```
“EzSystems\\EzPlatformSolrSearchEngine\\Tests\\SetupFactory\\”: “tests/lib/SetupFactory”
```

alongosz:
> why not just add something like the following to autoload array so kernel can stay the same across all versions:
> ```
> “EzSystems\\EzPlatformSolrSearchEngine\\Tests\\SetupFactory\\”: “tests/lib/SetupFactory”
> ```

Ok. Didn't want to do this at first to keep Tests in autoload-dev, but this is really the simplest solution and does not differ much in concept than moving `SetupFactory`. Thanks :+1: 

Cleaned up and squashed commits, ready for final approval.
I also kept here improvement of `LegacySetupFactory` itself, which removes dependency on kernels `config.php` file location.

All is green now here and also in ezsystems/ezpublish-kernel#2093 (however in the Kernel PR there's still TMP commit pointing to this branch, so this PR needs to be merged first).

alongosz:
>  but needs to go into 1.4 branch 

@andrerom done, sorry, sometimes I don't know where my head is :)

andrerom:
Thanks, merged :)

PR #104, opened 2017-10-02
EZP-28005: Use POST request for /select resource

pspanja:
> issue: https://jira.ez.no/browse/EZP-28005

This switches from `GET` to `POST` when requesting Solr `/select` endpoint.

In some cases very long queries are generated by the engine, hitting the server limit and causing errors. It can happen for example with complex user permissions or having lots of conditions on a field whose name is used in multiple ContentTypes.

While `GET` would be formally correct for `/select` resource, `POST` is more practical since it's size limit is higher. For example with Jetty `GET` is limited to 8k and `POST` to 200k.

Other than that it makes no difference for Solr and multiple projects already did the same for the same reason:

- https://github.com/lucidworks/banana/pull/159
- https://www.drupal.org/node/761990

### DIFF
--- a/lib/Gateway/Native.php
+++ b/lib/Gateway/Native.php
@@ -545,11 +545,17 @@ class Native extends Gateway
         $queryString = $this->generateQueryString($parameters);
 
         $response = $this->client->request(
-            'GET',
+            'POST',
             $this->endpointRegistry->getEndpoint(
                 $this->endpointResolver->getEntryEndpoint()
             ),
-            "/select?{$queryString}"
+            '/select',
+            new Message(
+                [
+                    'Content-Type' => 'application/x-www-form-urlencoded',
+                ],
+                $queryString
+            )
         );
 
         // @todo: Error handling?


emodric:
Does it make sense to backport this to previous versions (1.3 & 1.4), instead of targeting `master`?

pspanja:
@emodric I think so, shouldn't be a issue.

andrerom:
Yes, 1.4, as well as 1.0 is still maintained. so if possible target 1.0 here even.

pspanja:
Changed to 1.0 now, I have no idea if it will test.

andrerom:
Works, could you also create a bug report for this? for change log

volarname:
hello, when it will be merged? dont forget to older versions ;)

andrerom:
soon :)

andrerom:
Tagged 1.0.9: https://github.com/ezsystems/ezplatform-solr-search-engine/releases/tag/v1.0.9

For eZ Platform 1.7+ users it is already available as part of:
https://github.com/ezsystems/ezplatform-solr-search-engine/releases/tag/v1.4.2.1

PR #105, opened 2017-10-12
EZP-28064: Add FieldLike Criterion support

wizhippo:


lib/Query/Common/CriterionVisitor/Field/FieldLike.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\EzPlatformSolrSearchEngine\Query\Common\CriterionVisitor\Field;
```
alongosz: Hmm, I missed one: there should be no blank line before `namespace` :( Sorry for the inconvenience.

alongosz:
Thanks @wizhippo :)
Note: ezrobot review to be disregarded, checked manually.

andrerom:
@wizhippo you can now rebase on top of master to double check we got all cs things right.

andrerom:
@pspanja Seems the Fulltext tests fails on Solr 4, known? Should we just skip those tests in that case?

andrerom:
Ok, issue is tests assumes scoring works on location search when they don't do that on Solr 4. So this is all good, merging, thanks @wizhippo!

PR #106, opened 2017-10-12
Allow use of RegExp in FieldIn Common CriteriaVisitor

antoligy:


antoligy:
Above feedback isn't relevant within the scope of this PR.

alongosz:
no worries, just a cs-fixer update. I'll let you know when I fix it :)

andrerom:
Regarding this PR I'm unsure, it seems to add a feature not available on the SQL engine, so better would be to make something that can work across both for the use case here.

antoligy:
I'm not sure that that's entirely valid: Currently, it's possible to do the exact opposite with the SQL engine, i.e. insert SQL wildcards.

andrerom:
Ok, we'll try to investigate that with integration tests on API to better define this, thanks for finding this.

andrerom:
@alongosz We'll need to check from API perspective what we want to do with this one, for instance if we want to limit use to % to LIKE across engines.

antoligy:
exciting times!!

PR #109, opened 2017-10-17
EZP-28088: Re-try connections when trying to reach Solr

andrerom:
> issue: https://jira.ez.no/browse/EZP-28088

Note: This does not handle HTTP error codes returned, as code uses `ignore_errors`, however afaik what is relevant here is cases where we don't reach Solr _(network errors/dns/..)_. But if anyone see any other cases that would be relevant here please share :)

Also open question is if Solr would sometimes needs more time than 10s for handling calls, so might be timeout is too low here.

Todo:
- [x] Logger

### DIFF
--- a/lib/Gateway/HttpClient/Stream.php
+++ b/lib/Gateway/HttpClient/Stream.php
@@ -13,12 +13,18 @@ namespace EzSystems\EzPlatformSolrSearchEngine\Gateway\HttpClient;
 use EzSystems\EzPlatformSolrSearchEngine\Gateway\HttpClient;
 use EzSystems\EzPlatformSolrSearchEngine\Gateway\Message;
 use EzSystems\EzPlatformSolrSearchEngine\Gateway\Endpoint;
+use Psr\Log\LoggerInterface;
 
 /**
  * Simple PHP stream based HTTP client.
  */
 class Stream implements HttpClient
 {
+    /**
+     * @var \Psr\Log\LoggerInterface
+     */
+    private $logger;
+
     /**
      * @var int
      */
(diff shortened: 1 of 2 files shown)


lib/Gateway/HttpClient/Stream.php
```diff
@@ -34,12 +34,32 @@ class Stream implements HttpClient
...
+        $i = 0;
+        do {
+            ++$i;
+            if ($responseMessage = $this->requestStream($method, $endpoint, $path, $message)) {
+                return $responseMessage;
+            }
+
+            // Wait for 100ms before we retry
+            // Timeout is 10s, so time spent in worst case is 50.5s, which is less then default_socket_timeout (60s)
+            usleep(100000);
```
adamwojs: Extract hardcoded values to const would make the code more self-explanatory and easier to configure IMHO. 

```php
// Connection timeout in ms
const CONNECTION_TIMEOUT = 10000; 
// Number of connection attempts
const CONNECTION_RETRY_TIMES = 5;
// Connection retry in ms. Timeout is 10s, so time spent in worst case is 50.5s, 
// which is less then default_socket_timeout (60s)
const CONNECTION_RETRY_DELAY = 100;

// ...

for ($i = 0; $i < self::CONNECTION_RETRY_TIMES; $i++) {
	$responseMessage = $this->requestStream($method, $endpoint, $path, $message)
	if (responseMessage !== null) {
    	    return $responseMessage;
	}

 	// Wait before we retry
 	usleep(1000 * self::CONNECTION_RETRY_DELAY);
}
```

andrerom: true, but impossible yet to set them as private const, so might be better with properties?

adamwojs: +1, properties might be better.

andrerom:
@adamwojs Better?

PR #110, opened 2017-10-17
[WIP] Rip out sharding logic to be able to work on platform.sh

amirkoklan:
On platform.sh, setting of the shards makes the whole solr broken as the solr server's DNS cannot resolve the shards host.

that is how it works for my case, do not hesitate to reject the PR if you have a better way of solving it.

### DIFF
--- a/lib/Gateway/Native.php
+++ b/lib/Gateway/Native.php
@@ -129,22 +129,12 @@ class Native extends Gateway
      */
     protected function internalFind(array $parameters, array $languageSettings = array())
     {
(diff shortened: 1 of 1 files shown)


andrerom:
We don't currently have any other patch, however this one breaks the tests so won't be able to accept it as is either.

amirkoklan:
Thanks! you can assign that ticket to me. I'll try to finish the tests until next week.

andrerom:
No need @amirkoklan, the whole patch needs to be done differently. Basically shard variables needs to be set if there are shards involved, but when there is not this variable should not be applied.

PR #113, opened 2017-11-02
EZP-28176: Support for parallel and selective indexing command

andrerom:
> issue: https://jira.ez.no/browse/EZP-28176
> Depends on https://github.com/ezsystems/ezpublish-kernel/pull/2139

As a Developer I want a indexing command which is:
- faster at indexing (parallel indexing)
- able to refresh parts of index (by date or content id)
- able to omit purging so index is gradually refreshed on running prod setup
- able to remove index items if no longer existing (when I provide content id's to index)

Todo:
- [x] Split out SolrCreateIndexCommand.php changes into own PR and improvement for 1.0 and up
- [x] Re start travis once kernel PR has been merged

### DIFF
--- a/lib/Indexer.php
+++ b/lib/Indexer.php
@@ -9,83 +9,62 @@
 namespace EzSystems\EzPlatformSolrSearchEngine;
 
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
-use eZ\Publish\Core\Search\Common\Indexer as SearchIndexer;
-use EzSystems\EzPlatformSolrSearchEngine\Handler as SearchHandler;
-use Symfony\Component\Console\Helper\ProgressBar;
-use Symfony\Component\Console\Output\OutputInterface;
-use PDO;
-use RuntimeException;
+use eZ\Publish\Core\Persistence\Database\DatabaseHandler;
+use eZ\Publish\Core\Search\Common\IncrementalIndexer;
+use EzSystems\EzPlatformSolrSearchEngine\Handler as SolrSearchHandler;
+use eZ\Publish\SPI\Persistence\Handler as PersistenceHandler;
+use Psr\Log\LoggerInterface;
 
-class Indexer extends SearchIndexer
+class Indexer extends IncrementalIndexer
 {
     /**
      * @var \EzSystems\EzPlatformSolrSearchEngine\Handler
      */
     protected $searchHandler;
 
-    /**
-     * Create search engine index.
-     *
-     * @param \Symfony\Component\Console\Output\OutputInterface $output
-     * @param int $iterationCount
-     * @param bool $commit
-     */
-    public function createSearchIndex(OutputInterface $output, $iterationCount, $commit)
-    {
-        $output->writeln('Creating Solr Search Engine Index...');
+    public function __construct(
+        LoggerInterface $logger,
+        PersistenceHandler $persistenceHandler,
+        DatabaseHandler $databaseHandler,
+        SolrSearchHandler $searchHandler
+    ) {
+        parent::__construct($logger, $persistenceHandler, $databaseHandler, $searchHandler);
+    }
 
-        if (!$this->searchHandler instanceof SearchHandler) {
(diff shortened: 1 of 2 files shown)


MarioBlazek:
@andrerom did you maybe thought about adding option for refreshing index for subtree  ?

andrerom:
@MarioBlazek No, but that's a good idea, things like that is very simple to add here now. But I'll definitely add that one asap ;)

MarioBlazek:
@andrerom we've implemented that on current project. I can do PR on kernel.

andrerom:
@MarioBlazek If so then do the PR on top of  ezsystems/ezpublish-kernel#2139, you'll only need a new `getStatement*` method and handle the console option same way as `since.`

lib/Indexer.php
```diff
@@ -9,83 +9,68 @@
...
+        foreach ($contentIds as $contentId) {
+            try {
+                $info = $contentHandler->loadContentInfo($contentId);
+                if ($info->isPublished) {
+                    $content = $contentHandler->load($contentId, $info->currentVersionNo);
                     $documents[] = $this->searchHandler->generateDocument($content);
-                } catch (NotFoundException $e) {
-                    // Ignore content objects that have some sort of missing data on it
-                    $this->logWarning($progress, 'Content with id ' . $content->versionInfo->id . ' has missing data, so skipped for indexing. Full exception: ' . $e->getMessage());
+                    continue;
```
adamwojs: if we eliminate `continue` statement code will be more readable IMHO. 

```php
try {
    $info = $contentHandler->loadContentInfo($contentId);
    if ($info->isPublished) {
        $documents[] = $this->searchHandler->generateDocument(
            $contentHandler->load($contentId, $info->currentVersionNo)
        );
    }
} catch(NotFoundException $e) {
    // Catch this so we delete the index for this content below 
    $this->searchHandler->deleteContent($contentId);
}
```

lib/Indexer.php
```diff
@@ -9,83 +9,68 @@
...
-
-        $stmt = $this->getContentDbFieldsStmt(['count(id)']);
-        $totalCount = (int) $stmt->fetchColumn();
-        $stmt = $this->getContentDbFieldsStmt(['id', 'current_version']);
+        return 'eZ Platform Solr Search Engine';
+    }
 
+    public function purge()
+    {
+        $this->checkSearchEngine();
```
adamwojs: Do we need to check if `searchHandler` is supported in each call of `purge` and `updateSearchIndex`?

andrerom: alternative is that I overload ctor

lib/Indexer.php
andrerom: you are forgetting the non published code, it would mean:

```php
try {
    $info = $contentHandler->loadContentInfo($contentId);
    if ($info->isPublished) {
        $documents[] = $this->searchHandler->generateDocument(
            $contentHandler->load($contentId, $info->currentVersionNo)
        );
    } else {
            $this->searchHandler->deleteContent($contentId);
    }
} catch(NotFoundException $e) {
    $this->searchHandler->deleteContent($contentId);
}
```

prefer this? (I'm agnostic to it tbh, but I can see this is probably slightly easier to read)

adamwojs: Yes, I prefer this version because it explicitly says when content is added/removed from the search index. But taking into account the forgotten case, we can stay as is 😉

lib/Indexer.php
adamwojs: Overloading the constructor is OK for me, let's keep it simple. WDYT @alongosz?

alongosz: > Overloading the constructor is OK for me, let's keep it simple.

+1, makes sense

lib/Indexer.php
andrerom: @alongosz Any preferences here?

andrerom: This and constructor suggestion fixed in 8a27d12

Also fixed both in kernel PR

alongosz: current version looks ok

PR #117, opened 2018-02-14
EZP-28814: Make single core Solr work on platform.sh

andrerom:
> Replaces #110 
> Attempt to fix usage on platform.sh where Solr does not know about dns of itself causing it to break

This is kind of working around the platform.sh dns issue _(solr service does not known about it's own dns)_, and it's only able to do it on single core, multi core is still broken on Platform.sh which means any serious multi language setup _(where you want to configure different language stemming rules and so on per core)_ will be blocked from using that.

####  The real solution to this
- Platform.sh needs to fix their self discovery of their Solr setup, they anyway need that if people are to use multi / shared core setup.
- Far stretch, but might also be that this problem goes away if we enable solr cloud features, see pr on that.

/cc @vidarl

### DIFF
--- a/lib/Gateway/EndpointRegistry.php
+++ b/lib/Gateway/EndpointRegistry.php
@@ -62,4 +62,18 @@ class EndpointRegistry
 
         return $this->endpoint[$name];
     }
+
+    /**
+     * Get first Endpoint, for usecases where there is only one.
+     *
+     * @return \EzSystems\EzPlatformSolrSearchEngine\Gateway\Endpoint
+     */
+    public function getFirstEndpoint()
+    {
+        if (empty($this->endpoint)) {
+            throw new OutOfBoundsException("No Endpoint registered at all'.");
+        }
+
+        return reset($this->endpoint);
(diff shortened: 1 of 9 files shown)


nmeirik:
Great news! Looking forward to checking this out in the next release.

andrerom:
@nmeirik Already out in `v1.5.1.1` tag ;) _(PS: @vidarl will look into documenting how to use it, as you might know you need to copy and adapt the config into the root repo so platform.sh can use the solr config)_

nmeirik:
@andrerom You mean like [this](https://github.com/janit/ez-platform-cloud-solr-config/tree/master/.platform/solr)?

nmeirik:
@andrerom And also, us bleeding edge eZ guys are of course using eZ Platform 2.x. ;)

andrerom:
> @andrerom You mean like this?

yes

> @andrerom And also, us bleeding edge eZ guys are of course using eZ Platform 2.x. ;)

But of-course ;)

PR #118, opened 2018-03-06
EZP-28898: Implemented SortClause visitor for Random sort

wizhippo:
| Question           | Answer
| **JIRA issue**     | [EZP-28898](https://jira.ez.no/browse/EZP-28898)
| **Requires** | ezsystems/ezpublish-kernel#2605
| **Bug/Improvement**| no
| **New feature**    | yes
| **Target version** | Solr Bundle `2.0` for eZ Platform `3.0`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | yes

This allows random sorting based on a seed.  Using the same seed will result in same order of random results.

Example:

This example rotates the used seed based on time so the returned results have a new order every 360 seconds.

```
$query = new Query();
$query->sortClauses = [
    new Random((int)(time() / 360)),
];

```

### DIFF
--- a/lib/Query/Common/SortClauseVisitor/Random.php
+++ b/lib/Query/Common/SortClauseVisitor/Random.php
@@ -37,6 +37,7 @@ class Random extends SortClauseVisitor
      */
     public function visit(SortClause $sortClause)
     {
-        return 'random_' . $sortClause->targetData->seed . $this->getDirection($sortClause);
+        $seed = $sortClause->targetData->seed ?? mt_rand();
+        return 'random_' . (string)$seed . ' ' . $this->getDirection($sortClause);
     }
 }
--- a/lib/Query/Common/SortClauseVisitor/Random.php
+++ b/lib/Query/Common/SortClauseVisitor/Random.php
@@ -1,0 +1,42 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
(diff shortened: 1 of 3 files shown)


Plopix:
That is kind of cool!

lib/Query/Common/SortClauseVisitor/Random.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
```
alongosz: Please drop this line, unsupported for many years now ;)

wizhippo:
@alongosz Updated PR description.  Let me know if you are looking for more or if it is unclear.

alongosz:
> @alongosz Updated PR description. Let me know if you are looking for more or if it is unclear.

Thanks! This is what I was looking for. We'll try to wrap this up in the next one or two weeks.

andrerom:
@alongosz Are we able to support such a seed across all Search engines?

alongosz:
> @alongosz Are we able to support such a seed across all Search engines?

For now it seems yes, seed is usually `int`, for LSE on PostgreSQL it is float <-1, 1>, but this can be interpolated/normalized. By default it won't be required (`?int`). Ref: https://github.com/ezsystems/ezpublish-kernel/pull/2605#discussion_r283246656

_This could be solved by some implementation-independent structures, but right now I don't have a clear picture of such solution._

alongosz:
This will be a part of eZ Platform `3.0.0-beta1`. Thank you @wizhippo and @ViniTou :tada:

PR #120, opened 2018-03-21
EZEE-1903: Added bin/generate-solr-config.sh

vidarl:
This is is a small script that generates the solr config. 

This is particular useful if using Platform.sh has they require that config to be present in your source tree

Also, see https://github.com/ezsystems/ezplatform/pull/282

### DIFF
--- a/bin/.travis/init_solr.sh
+++ b/bin/.travis/init_solr.sh
@@ -10,7 +10,7 @@ default_cores[2]='core2'
 default_cores[3]='core3'
 
 SOLR_PORT=${SOLR_PORT:-8983}
-SOLR_VERSION=${SOLR_VERSION:-6.6.0}
+SOLR_VERSION=${SOLR_VERSION:-6.6.5}
 SOLR_DEBUG=${SOLR_DEBUG:-false}
 SOLR_HOME=${SOLR_HOME:-ez}
 SOLR_CONFIG=${SOLR_CONFIG:-${default_config_files[*]}}
@@ -21,7 +21,7 @@ SOLR_INSTALL_DIR="${SOLR_DIR}/${SOLR_VERSION}"
 download() {
     case ${SOLR_VERSION} in
         # PS!!: Append versions and don't remove old once, kernel uses this script!
-        4.10.4|6.3.0|6.4.1|6.4.2|6.5.1|6.6.0 )
+        4.10.4|6.3.0|6.4.1|6.4.2|6.5.1|6.6.0|6.6.5 )
             url="http://archive.apache.org/dist/lucene/solr/${SOLR_VERSION}/solr-${SOLR_VERSION}.tgz"
             ;;
         *)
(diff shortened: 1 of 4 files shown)


andrerom:
As long as we don't use it for any automation that can be covered by tests, I'm unsure if I see a real need for this. Afaik devs tend to modify solr config and use that for projects.

andrerom:
@alongosz failure here is unrelated, are you ok with adding this to simplify doc? @MagdalenaZuba is working on some doc for Solr usage.

bin/generate-solr-config.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+
+set -e
+
+DESTINATION_DIR=.platform/configsets/solr6/conf
```
alongosz: Since this is clearly for Platform.sh-only, the script name `generate-solr-config.sh` is a bit misleading, so please either:
1) make it generic (not sure if this makes sense in this case though)
2) rename it to something like: `platformsh-generate-solr-config.sh` or `generate-platformsh-solr-config.sh` or to other meaningful name ;)

andrerom: As far as I read the Readme change this can also be used for generic needs, using the `--destination-dir` option.
This is one of the reason why I came back to this as it can thus simplify doc a bit as well.

andrerom: Best would be if we could find a way to use this for integration tests when it generates config so this gets constant testing.

+ then also update Readme for how to setup project for testing.

andrerom:
> So +1 but please, let's do QA on that ;)

yes, I agree. Can be done as part of Doc story.

andrerom:
@alongosz Updated to make Travis use this script for Solr 6, does your review still stand?

andrerom:
@micszo Yes, sorry we probably should have done doc first, changes on .platform config always need s to be done before you push your changes _(aka deploy)_.

Process is:
1. Run this command on your local code checkout, and as last output of command suggests adapt .platform.app.yml and .platform/ config.
2. commit and push

micszo:
OK, I ran the command locally (tried before and after adapting the configs) and now it fails with error:
`cp: ./vendor/ezsystems/ezplatform-solr-search-engine/bin/../../../vendor/ezsystems/ezplatform-solr-search-engine/lib/Resources/config/solr/*: No such file or directory`
screen shot 2018-07-24 at 15 04 07

andrerom:
@micszo Should be fixed now, but as @vidarl is back next week we should let him do a pass on this to make sure it works as he intended, and also verify it works for Linux. So you can postpone looking at this until next week.

@vidarl Changes here besides some cleanups are to accomplish two things: Work on Mac Os, and change so our travis scripts can reuse at least parts of this so it gets some continues testing.

bin/generate-solr-config.sh
```diff
@@ -0,0 +1,134 @@
...
+            ;;
+        *)
+            show_help "${i}"
+            exit 1
+            ;;
+    esac
+done
+
+
+if [ `whoami` == "root" ]; then
```
andrerom: note: maybe we should add a `--allow-root` argument for use on docker and such.

andrerom:
@vidarl Could you review this and see if the adaptions looks ok for linux?

micszo:
Hi. I gave it another try on Mac with the latest commit and this time it finished but with warnings.
Could you have a look?
screen shot 2018-08-03 at 09 50 21

andrerom:
@micszo Fixed, to used to writing markdown 😊

vidarl:
Looks good on linux too.

We should not forget about https://github.com/ezsystems/ezplatform/pull/282 when merging this...

andrerom:
@vidarl do you take care about merges? I guess this one needs to go first, and we should probably make a release here with this before merging the meta PR.

vidarl:
Merged and created release.

Also merged https://github.com/ezsystems/ezplatform/pull/282

alongosz:
@vidarl now kernel tests for Solr fail: https://travis-ci.org/ezsystems/ezpublish-kernel/jobs/413939608#L767 :(
I tried to restart it, but the problem doesn't go away.

bin/.travis/init_solr.sh
```diff
@@ -221,7 +198,9 @@ create_core() {
 download
 
 if [[ ${SOLR_VERSION} == 6* ]] ; then
-    configure
+    ./bin/generate-solr-config.sh \
```
alongosz: I missed that, we cannot assume that our CWD always would be relative to that path. We should rather make it relative to this (init_solr.sh) script - if possible.

vidarl: right.. I'll look into it

vidarl: Have a look : https://github.com/ezsystems/ezplatform-solr-search-engine/pull/125

PR #122, opened 2018-05-22
EZP-28739: Add --continue-on-error option & display corrupted content IDs in the search indexer

kmadejski:
…t IDs in the search indexer

JIRA issue: https://jira.ez.no/browse/EZP-28739

This PR contains changed required for changes introduced in: https://github.com/ezsystems/ezpublish-kernel/pull/2221

### DIFF
--- a/lib/Indexer.php
+++ b/lib/Indexer.php
@@ -23,11 +23,6 @@ class Indexer extends IncrementalIndexer
      */
     protected $searchHandler;
 
-    /**
-     * @var \Psr\Log\LoggerInterface
-     */
-    protected $logger;
-
     public function __construct(
         LoggerInterface $logger,
         PersistenceHandler $persistenceHandler,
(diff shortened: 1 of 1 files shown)


lib/Indexer.php
```diff
@@ -22,13 +23,19 @@ class Indexer extends IncrementalIndexer
...
+    protected $logger;
+
     public function __construct(
         LoggerInterface $logger,
         PersistenceHandler $persistenceHandler,
         DatabaseHandler $databaseHandler,
         SolrSearchHandler $searchHandler
     ) {
         parent::__construct($logger, $persistenceHandler, $databaseHandler, $searchHandler);
+        $this->logger = $logger;
```
andrerom: Why is this needed? Seems to be `protected` in Indexer abstract, so we should be able to read it here right?

kmadejski: 🤦‍♂️  - fixed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/122/commits/057fa745751c88360faa3ff8fd600844a41c876d

adamwojs (approved):

konradoboza (approved):
Option's name is `continue-on-error` so the titles of PRs and ticket are a bit confusing. :)

PR #123, opened 2018-07-11
EZP-29396: Remove unnecessary test cache decorator

kmadejski:
| Question           | Answer
| **JIRA issue**     | [EZP-29396](https://jira.ez.no/browse/EZP-29396)
| **Bug**| yes
| **Target version** | `1.5` and `master`

This change is required regarding changes done in kernel in: https://github.com/ezsystems/ezpublish-kernel/pull/2387.
Since test should be working properly with Redis, using of the `eZ\Publish\Core\Persistence\Cache\Tests\Helpers\IntegrationTestCacheServiceDecorator` is no longer necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,11 +18,11 @@ matrix:
           env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="4.10.4" CORES_SETUP="shared"
         - php: 7.1
           env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="4.10.4" CORES_SETUP="single" SOLR_CORES="collection1"
-        - php: 7.0
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="6.4.2"  CORES_SETUP="dedicated"
-        - php: 7.1
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="6.5.1"  CORES_SETUP="shared" COMPOSER_REQUIRE="ezsystems/ezpublish-kernel:~6.7.4@dev"
         - php: 5.6
+          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="6.6.0"  CORES_SETUP="single" SOLR_CORES="collection1"
+        - php: 7.0
+          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="6.4.2"  CORES_SETUP="dedicated" COMPOSER_REQUIRE="ezsystems/ezpublish-kernel:~6.7.4@dev"
+        - php: 7.1
           env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="6.6.0"  CORES_SETUP="single" SOLR_CORES="collection1"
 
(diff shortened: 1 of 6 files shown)


andrerom:
@kmadejski Seems we have issue from newer version og phpunit packages on PHP 7.1 and higher, could you check?

ref:
```
PHP Fatal error:  Class 'PHPUnit\Framework\Constraint\Constraint' not found in /home/travis/build/ezsystems/ezplatform-solr-search-engine/vendor/ezsystems/ezpublish-kernel/eZ/Publish/API/Repository/Tests/PHPUnitConstraint/ValidationErrorOccurs.php on line 19
```

kmadejski:
@andrerom yes, actually, I'm investigating it right now.

alongosz:
@kmadejski we just need to upgrade PHPUnit to version `^7.0` because we use Kernel Setup Factory for integration tests, which uses new version of PHPUnit. Composer does not catch this, because it's `require-dev` package.

kmadejski:
@alongosz correct 🙂 (https://github.com/ezsystems/ezplatform-solr-search-engine/pull/123/commits/19bd8cd7ea5a1f5d067c4c56c67a3bf44f7f3159). Let's see whether the test will fail again.

.travis.yml
```diff
@@ -18,13 +18,13 @@ matrix:
...
         - php: 7.1
           env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="4.10.4" CORES_SETUP="single" SOLR_CORES="collection1"
-        - php: 7.0
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="6.4.2"  CORES_SETUP="dedicated"
-        - php: 7.1
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="6.5.1"  CORES_SETUP="shared" COMPOSER_REQUIRE="ezsystems/ezpublish-kernel:~6.7.4@dev"
         - php: 5.6
           env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="6.6.0"  CORES_SETUP="single" SOLR_CORES="collection1"
-
+        - php: 5.6
```
andrerom: Coverage for 7.1 on Solr 6 is missing now

maybe also just move `COMPOSER_REQUIRE="ezsystems/ezpublish-kernel:~6.7.4@dev"` to 7.0 job to avoid the shuffling of the list here also.

kmadejski: > Coverage for 7.1 on Solr 6 is missing now

Right, I'll add this, but this can be only against `ezpublish-kernel` master.

> maybe also just move COMPOSER_REQUIRE="ezsystems/ezpublish-kernel:~6.7.4@dev" to 7.0

Unfortunately, we can't run test with PHP >= 7.0 and `ezsystems/ezpublish-kernel:~6.7.4@dev`  because of the problem with typehints for return values in the new PHPUnit (ref: https://travis-ci.org/ezsystems/ezplatform-solr-search-engine/jobs/403021480).

andrerom: but PHPUnit 7.x requires PHP 7.1 doesn't it? ref: https://phpunit.de/announcements/phpunit-7.html

kmadejski: Ah, you are right. Moved tests against 6.7 to PHP 7.0.

PR #126, opened 2018-08-13
Test regressions

pspanja:
Testing change in fulltext scoring probably caused by https://github.com/ezsystems/ezplatform-solr-search-engine/pull/120 (not sure why it didn't show initially).

Discovered on https://github.com/netgen/ezplatform-search-extra/pull/7, see the change in regression tests between last two commits (last build in the list), installing `1.5.4` and `1.5.2.1` respectively:

- https://travis-ci.org/netgen/ezplatform-search-extra/jobs/415284876
- https://travis-ci.org/netgen/ezplatform-search-extra/jobs/415293559

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,22 +10,8 @@ cache:
 
 matrix:
     include:
-        - php: 5.6
-          env: TEST_CONFIG="phpunit.xml"
         - php: 7.2
-          env: TEST_CONFIG="phpunit.xml"
-        - php: 5.6
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="4.10.4" CORES_SETUP="dedicated" COMPOSER_REQUIRE="ezsystems/ezpublish-kernel:~6.7.4@dev"
-        - php: 7.0
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="4.10.4" CORES_SETUP="shared"
-        - php: 7.1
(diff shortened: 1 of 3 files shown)


pspanja:
OK, it doesn't want to fail here. I need to do some more digging :)

andrerom:
Related change: https://github.com/ezsystems/ezplatform-solr-search-engine/commit/b9a8860908c4a53c05da4a4ab5bb27d1cdefe440

pspanja:
The cause is definitely `autoSoftCommit` change, but I'm getting completely inconsistent behavior.
I'm going to try to provoke the error here again.

andrerom:
Ah, this one? https://github.com/ezsystems/ezpublish-kernel/pull/2416 In regards to expected size of sub group count right?

pspanja:
It was related to fulltext search, but it suddenly doesn't fail anymore, and I have no answer why :)

Re ezsystems/ezpublish-kernel#2416 it actually makes sense, since sub group count uses search: https://github.com/ezsystems/ezpublish-kernel/blob/6.7/eZ/Publish/Core/Repository/UserService.php#L207

pspanja:
Ah, right :) On https://github.com/netgen/ezplatform-search-extra/pull/7 I kind of proved the failures are caused by https://github.com/ezsystems/ezplatform-solr-search-engine/pull/120, but I still have no idea about the cause and why it doesn't reproduce here. The weirdest thing is that it apparently randomly passes.

I'll just go with local previous version of `init_solr.sh`, since it fixes the problem for now.

PR #128, opened 2019-01-04
EZP-29984: Add support for Field LIKE criterion with "*" wildcards

andrerom:
https://jira.ez.no/browse/EZP-29984

Adapt Solr Search Engine to correct support for LIKE with wildcards using "*".

Todo:
- [x] kernel PR: https://github.com/ezsystems/ezpublish-kernel/pull/2517
- [x] CS / Adapt unit tests

### DIFF
--- a/lib/Query/Common/CriterionVisitor/Field/FieldIn.php
+++ b/lib/Query/Common/CriterionVisitor/Field/FieldIn.php
@@ -5,8 +5,6 @@
  *
  * @copyright Copyright (C) eZ Systems AS. All rights reserved.
  * @license For full copyright and license information view LICENSE file distributed with this source code.
- *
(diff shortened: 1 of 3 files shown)


wizhippo:
@andrerom I'm good with it like this.  Are we making legacy behave the same (using '%' on the start and end by default)?  If not then making this behave the same for consistency may be wanted.

andrerom:
@wizhippo It's on the todo on this. Before this can be merged, there should be PR on kernel adding integration tests and possibly adapt logic in legacySE.

andrerom:
@wizhippo [Kernel PR done](ezsystems/ezpublish-kernel#2517) and this PR adapted for it _(using * now instead to align with FullText)_.

PR #131, opened 2019-02-18
EZP-30169: Bump to 1.6 bundle version & drop Solr 4 support

andrerom:
Story: https://jira.ez.no/browse/EZP-30169

Preparations for the 2.5LTS, to reduce test/support matrix to get rid of irrelevant versions _(by now)_.

Given location search scoring can never be supported on Solr 4 and given how old it is, it makes sense to stop using time on it for QA and others.

### DIFF
--- a/bin/.travis/init_solr.sh
+++ b/bin/.travis/init_solr.sh
@@ -24,7 +24,7 @@ SCRIPT_DIR=`dirname $0`
 download() {
     case ${SOLR_VERSION} in
         # PS!!: Append versions and don't remove old once, kernel uses this script!
-        4.10.4|6.3.0|6.4.1|6.4.2|6.5.1|6.6.0|6.6.5 )
+        6.3.0|6.4.1|6.4.2|6.5.1|6.6.0|6.6.5 )
             url="http://archive.apache.org/dist/lucene/solr/${SOLR_VERSION}/solr-${SOLR_VERSION}.tgz"
             ;;
         *)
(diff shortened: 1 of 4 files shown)


bin/.travis/init_solr.sh
```diff
@@ -199,12 +143,12 @@ create_core() {
...
             --destination-dir="${SOLR_INSTALL_DIR}/server/${SOLR_HOME}/template"
     run
-else
-    solr4_configure
-    solr4_run
-fi
+#else
+#    solr4_configure
+#    solr4_run
+#fi
```
alongosz: nitpick: since you dropped these functions, I don't see a need to keep this entire if-else-if commented out, can be just dropped as well.

andrerom: we can, I was just anticipating that we need cusom logic again once we add solr 7 support.

andrerom:
> In theory we should probably bump the Bundle to `2.0`, but no strong feelings here ;)

In tend to agree also, but:
- looking at how Doctrine does things, they don't seem to agree bumping libs or service versions needs a major version, they does this also in feature releases
- If we do want to make a 2.0 we should strongly switch to Solr Cloud support as well (config BC break), but I don't think engineering has capacity to deal with that atm

bin/.travis/init_solr.sh
andrerom: fixed in ef4fbc0

PR #135, opened 2019-03-20
EZP-30282: Impl. indexing of related objects for the full text search

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-30282

## Description 

Impl. indexation of related objects for the full text search. 

### Indexing related content

Text of related content is stored in the `meta_related_content_X__text_t ` index fields where `X` is relation level e.g. `meta_related_content_1__text_t`, `meta_related_content_2__text_t ` 

### Boost factor for related content

As an User, I expect the following order of search results for query "Content A": `Content A`, `Content B`  when content repository contains `Content A` and  `Content B` with embedded `Content A`.
 
This is implemented by adding boost factor based on the following formula:

image

to the index fields with related content. Ref. 
https://github.com/ezsystems/ezplatform-solr-search-engine/blob/1238c57777b936c63a7b2d31a73032f8fee9e3c5/lib/Query/Content/CriterionVisitor/FullText.php#L154-L157
 
### Indexing depth configuration

As a Developer, you are able to define the maximum indexing depth using the following YAML configuration:

```yml
ez_search_engine_solr:
    # ...
    connections:
        default:
            # ...
            indexing_depth:
                default: 1          # Default value: 0 - no relation indexing, 1 - direct relations, 2nd level  relations, 3rd level  relations (maximum value).
                content_type:
                    article: 2      # Index depth defined for specyfic content type
``` 

Indexing of the related content is disabled by default, however, this might change in eZ Platform 3.X. 

#### Checklist:
- [x] Coding standards (`$ composer fix-cs`)
- [x] Impl. feature / bugifx
- [x] Impl. tests 
- [x] Ready for Code Review

### DIFF
--- a/lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentFulltextFields.php
+++ b/lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentFulltextFields.php
@@ -14,9 +14,11 @@ use eZ\Publish\Core\Search\Common\FieldNameGenerator;
 use eZ\Publish\Core\Search\Common\FieldRegistry;
 use eZ\Publish\SPI\Persistence\Content;
 use eZ\Publish\SPI\Persistence\Content\Type as ContentType;
+use eZ\Publish\SPI\Persistence\Content\Handler as ContentHandler;
 use eZ\Publish\SPI\Persistence\Content\Type\Handler as ContentTypeHandler;
 use eZ\Publish\SPI\Search\Field;
 use eZ\Publish\SPI\Search\FieldType;
+use EzSystems\EzPlatformSolrSearchEngine\FieldMapper\IndexingDepthProvider;
 
 /**
  * Maps Content fulltext fields to Content document.
@@ -31,9 +33,21 @@ class ContentDocumentFulltextFields extends ContentTranslationFieldMapper
     private static $fieldName = 'meta_content__text';
 
     /**
+     * Field of related content name, untyped.
+     *
+     * @var string
+     */
+    private static $relatedContentFieldName = 'meta_related_content_%d__text';
+
+    /**
      * @var \eZ\Publish\SPI\Persistence\Content\Type\Handler
      */
     protected $contentTypeHandler;
+
+    /**
+     * @var \eZ\Publish\SPI\Persistence\Content\Handler
+     */
+    protected $contentHandler;
 
     /**
      * @var \eZ\Publish\Core\Search\Common\FieldRegistry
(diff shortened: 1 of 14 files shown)


andrerom:
On the level thing, I'm unsure I would expect it to go past one level (by default). Thinking of this from a Page level as an example, what you want is perhaps to get some relevance for what "is" on the page from an editorial / business view, at least as a good default. And  as you suggest it should have lower boost by default then direct field content*.

\* But maybe it should be possible and documented how to can compensate this by boosting the field holding the relation, however to do that we need to rather move to index each field separately I think _(per discussion with @kmadejski and @pspanja on fixing index time boosting on fields)_.

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentFulltextFields.php
```diff
@@ -103,21 +155,77 @@ public function mapFields(Content $content, $languageCode)
...
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
+    private function doMapRelatedFields(Content $sourceContent, $languageCode, $maxDepth, $depth)
+    {
+        $fields = [];
+
+        $relations = $this->contentHandler->loadRelations($sourceContent->versionInfo->contentInfo->id);
+        foreach ($relations as $relation) {
+            $content = $this->contentHandler->load($relation->destinationContentId);
```
andrerom: Can be bulk loaded this? ;)

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentFulltextFields.php
```diff
@@ -103,21 +155,77 @@ public function mapFields(Content $content, $languageCode)
...
+            $contentType = $this->contentTypeHandler->load(
```
andrerom: Ideally also this, but as it is placed in in-memory cache now this one is not as critical, but you still end up with more code execution so.. ideally also bulk load this one based on the bulk loaded content list.

andrerom:
_New comment after looking at the updated diff_ 😄

Looks promising 👍 
Would it make sense to only allow depth of 0-3? Also it seems we might need some recursion protection here _(content1 -> content2 -> content1 or more advanccircularar relations, mainly skipping content1 being indexes several times in the index)_

andrerom:
Somewhat side here:

I'm wondering if this is maybe overlapping with the index boosting issue we have for fields that is being discussed with @kmadejski and @pspanja.

As in to add context on that; From my understand we either need to A. start to index all fields separately for fulltext in order to be able to boost them separately, or B. we need to consider using [Solr Payloads](https://lucidworks.com/2017/09/14/solr-payloads/) in order to be able to boost terms.

If A, then this stuff _could_ have been indexed as part of the relevant relation/relationList/page field full text index. But that said I realize it does not match needs for embeds on RichText which then will need either separate field or Payload as well... :)

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentFulltextFields.php
```diff
@@ -103,21 +155,86 @@ public function mapFields(Content $content, $languageCode)
...
+        }
+
+        return sprintf(self::$relatedContentFieldName, $depth);
+    }
+
     /**
      * Return index field type for the given $contentType.
      *
      * @param \eZ\Publish\SPI\Persistence\Content\Type $contentType
+     * @param $depth
```
kmadejski: Seems to be some copy-paste mistake 😉

PR #136, opened 2019-04-17
EZP-30835: Added support for Solr 7.7

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-30835

### Related SOLR issues

https://issues.apache.org/jira/browse/SOLR-11501:

> Starting a query string with local-params {!myparser ...} is used to switch the
query parser to another, and is intended for use by Solr system developers, not end users doing searches. To reduce negative side-effects of unintended hack-ability, we've limited the cases that local-params will be parsed to only contexts in which the default parser is "lucene" or "func". So if defType=edismax then q={!myparser ...} won't work. In that example, put the desired query parser into defType. Another example is if deftype=edismax then hl.q={!myparser ...} won't work for the same reason. In that example, either put the desired query parser into hl.qparser or set hl.qparser=lucene. Most users won't run into these cases but some will and must change. If you must have full backwards compatibility, use luceneMatchVersion=7.1.0 or something earlier.

Ref. https://github.com/ezsystems/ezplatform-solr-search-engine/pull/136/files#diff-d75c9ba9fe393b04edf2db8762aa77b7L65

https://issues.apache.org/jira/browse/SOLR-10503:

> CurrencyField has been deprecated in favor of new CurrencyFieldType

_Dropped as is not used by eZ Platform_

https://issues.apache.org/jira/browse/SOLR-10729:

> Deprecated LatLonType, GeoHashField, SpatialPointVectorFieldType, and SpatialTermQueryPrefixTreeFieldType. Instead, switch to LatLonPointSpatialField or SpatialRecursivePrefixTreeFieldType or RptWithGeometrySpatialField.

_Will be processed as a separate task: https://jira.ez.no/browse/EZP-30977_

https://issues.apache.org/jira/browse/SOLR-?????:

> Deprecated: All Trie* numeric and date field types have been deprecated in favor of *Point field types.

_Replaced according to suggestion_

https://issues.apache.org/jira/browse/SOLR-12770:

> The 'shards' parameter handling logic changes to use a new config element to determine what hosts can be requested.

_This is CI only related issue_

https://issues.apache.org/jira/browse/SOLR-?????:

> Index-time boosts have been removed from Lucene, and are no longer available from Solr. If any boosts are provided, they will be ignored by the indexing chain. As a replacement, index-time scoring factors should be indexed in a separate field and combined with the query score using a function query. See the section Function Queries for more information.

Ref. https://lucene.apache.org/solr/guide/7_2/major-changes-in-solr-7.html

### Release 

Due to the major changes in Solr, support for 7.X will be part of the `ezplatform-solr-search-engine` 2.0 release which will be compatible with eZ Platform 2.5 release. Solr 6.X will be dropped in this release. 

Solr 7.X and Solr 8.X will be officialy supported in `ezplatform-solr-search-engine` 3.0 release which will be compatible with eZ Platform 3.X.

##### TODO
- [X] Add entries to the test matrix
- [X] Adjust `bin/generate-solr-config.sh`
- [x] Adjust `schema.xml`

### DIFF
--- a/bin/.travis/init_solr.sh
+++ b/bin/.travis/init_solr.sh
@@ -15,7 +15,7 @@ default_shards=('shard0')
 
 SOLR_PORT=${SOLR_PORT:-8983}
 SOLR_DIR=${SOLR_DIR:-'__solr'}
-SOLR_VERSION=${SOLR_VERSION:-'6.6.5'}
+SOLR_VERSION=${SOLR_VERSION:-'7.7.2'}
 SOLR_INSTALL_DIR="${SOLR_DIR}/${SOLR_VERSION}"
 SOLR_DEBUG=${SOLR_DEBUG:-false}
 SOLR_HOME=${SOLR_HOME:-'ezcloud'}
@@ -44,7 +44,7 @@ fi
 download() {
     case ${SOLR_VERSION} in
         # PS!!: Append versions and don't remove old once, kernel uses this script!
-        6.3.0|6.4.1|6.4.2|6.5.1|6.6.0|6.6.5 )
+        7.7.2 )
             url="http://archive.apache.org/dist/lucene/solr/${SOLR_VERSION}/solr-${SOLR_VERSION}.tgz"
             ;;
         *)
@@ -131,19 +131,19 @@ wait_for_solr(){
     done
 }
 
-# Run for Solr 6
+# Run for Solr 7
 solr_run() {
     echo "Running with version ${SOLR_VERSION} in standalone mode"
     echo "Starting solr on port ${SOLR_PORT}..."
 
-    ./${SOLR_INSTALL_DIR}/bin/solr -p ${SOLR_PORT} -s ${SOLR_HOME} || exit_on_error "Can't start Solr"
+    ./${SOLR_INSTALL_DIR}/bin/solr -p ${SOLR_PORT} -s ${SOLR_HOME} -Dsolr.disable.shardsWhitelist=true || exit_on_error "Can't start Solr"
 
     echo "Started"
 
     solr_create_cores
 }
 
-# Create cores for Solr 6
+# Create cores for Solr 7
 solr_create_cores() {
     home_dir="${SOLR_INSTALL_DIR}/server/${SOLR_HOME}"
     template_dir="${home_dir}/template"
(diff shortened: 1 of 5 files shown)


andrerom:
Will this also pass on Solr 6.6 or we drop support for that in 1.6?

adamwojs:
> Will this also pass on Solr 6.6 or we drop support for that in 1.6?

I will check 😉  IMHO we should drop Solr 6.6 support in release with Solr 7. Changelog for Solr/Lucene between 6.6 and 7.X is big (includes BC breaks) which might make hard to have the support of the same feature set for both versions.

andrerom:
ok

bin/.travis/init_solr.sh
```diff
@@ -44,7 +44,7 @@ fi
 download() {
     case ${SOLR_VERSION} in
         # PS!!: Append versions and don't remove old once, kernel uses this script!
-        6.3.0|6.4.1|6.4.2|6.5.1|6.6.0|6.6.5 )
+        6.3.0|6.4.1|6.4.2|6.5.1|6.6.0|6.6.5|7.7.1 )
```
andrerom: Nitpick: 7.7.2 is out
Also, any idea how much they have broken in 8 in terms of also testing this branch on Solr 8.2?

adamwojs: > Nitpick: 7.7.2 is out 

Thanks for the information! I will update PR. 

> Also, any idea how much they have broken in 8 in terms of also testing this branch on Solr 8.2?

Let's add 8.2 to the test matrix and we will see 😉

bin/.travis/init_solr.sh
```diff
@@ -220,12 +220,15 @@ solr_cloud_configure_collection() {
     local files=("${SOLR_CONFIG[@]}")
     local config_dir="${INSTALL_DIR}/server/solr/configsets/basic_configs/conf"
 
-    files+=("${config_dir}/currency.xml")
     files+=("${config_dir}/stopwords.txt")
     files+=("${config_dir}/synonyms.txt")
     files+=("${config_dir}/elevate.xml")
     files+=("${config_dir}/solrconfig.xml")
 
+    if [[ ! $SOLR_VERSION =~ ^7 ]]; then
```
andrerom: should the logic be to only apply it if version is smaller than 7? (thinking about later steps to add 8 here)

andrerom:
@adamwojs we kind of need this in 2.5 as well. It seems we requires Debian 10 which had Java 11 which Solr 6 does not support :neckbeard:  That means rebasing this on 1.7 right?

adamwojs:
@andrerom I discussed with @lserwatka Solr 7+ support in eZ Platform 2.5 and we decided to go as described in "Release" section in the PR description. I hope it's fine for you.

bin/.travis/init_solr.sh
```diff
@@ -44,7 +44,7 @@ fi
 download() {
     case ${SOLR_VERSION} in
         # PS!!: Append versions and don't remove old once, kernel uses this script!
-        6.3.0|6.4.1|6.4.2|6.5.1|6.6.0|6.6.5 )
+        6.3.0|6.4.1|6.4.2|6.5.1|6.6.0|6.6.5|7.7.2 )
```
alongosz: If we decided to drop support for Solr `6.x` in `2.0` of this bundle, we probably should remove them here as well, to avoid suggestion we still support it.

bin/.travis/init_solr.sh
```diff
@@ -218,18 +218,28 @@ solr_cloud_configure_collection() {
     create_dir ${TEMPLATE_DIR}
 
     local files=("${SOLR_CONFIG[@]}")
-    local config_dir="${INSTALL_DIR}/server/solr/configsets/basic_configs/conf"
 
-    files+=("${config_dir}/currency.xml")
+    if [[ $SOLR_VERSION =~ ^(7) ]]; then
```
alongosz: Do we need this check if we support 7.x only now?

bin/.travis/init_solr.sh
```diff
@@ -218,18 +218,28 @@ solr_cloud_configure_collection() {
...
+
+    echo $config_dir;
+
+    files+=("${config_dir}/solrconfig.xml")
     files+=("${config_dir}/stopwords.txt")
     files+=("${config_dir}/synonyms.txt")
-    files+=("${config_dir}/elevate.xml")
-    files+=("${config_dir}/solrconfig.xml")
+
+    if [[ ! $SOLR_VERSION =~ ^(7) ]]; then
```
alongosz: same question here

adamwojs:
PR updated according to @alongosz code review suggestions.

composer.json
```diff
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": "^7.1",
-        "ezsystems/ezpublish-kernel": "^7.5@dev",
+        "ezsystems/ezpublish-kernel": "dev-solr_77 as 7.5",
```
alongosz: Friendly reminder ;)

m-tyrala:
Everything's fine for me, QA approve.

adamwojs:
PR has been rebased and squashed.

alongosz:
Requires ezsystems/ezpublish-kernel#2775 for tests to pass, so merging now.

alongosz:
@adamwojs could you merge up?

adamwojs:
Done.

PR #138, opened 2019-05-20
Clarify info message for the main_translations config node

adamwojs:
> JIRA: -

## Description

Clarify info message for the `main_translations` config node.

### DIFF
--- a/bundle/DependencyInjection/Configuration.php
+++ b/bundle/DependencyInjection/Configuration.php
@@ -244,7 +244,7 @@ class Configuration implements ConfigurationInterface
                                     'number of Solr endpoints that the query is ' .
                                     'distributed to when using always available fallback ' .
                                     "or searching only on the main languages.\n\n" .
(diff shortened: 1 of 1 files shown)


bundle/DependencyInjection/Configuration.php
```diff
@@ -243,7 +243,9 @@ function ($v) {
                                     'This setting is optional. Use it to reduce the ' .
                                     'number of Solr endpoints that the query is ' .
                                     'distributed to when using always available fallback ' .
-                                    'or searching only on the main languages.'
+                                    "or searching only on the main languages.\n\n" .
+                                    'Make sure main translations endpoint is not used in any other context, ' .
```
DominikaK:
```suggestion
                                    'Make sure the main_translations endpoint is not used in any other context, ' .
```

andrerom:
@adamwojs Is it ok if we cherry pick this to earlier branches to? Or is it not relevant there?

adamwojs:
@andrerom  Sure, it's valid for all supported branches.

PR #146, opened 2019-08-09
EZP-30846: Dropped deprecated Symfony\Component\Config\Definition\Builder\TreeBuilder::root method calls

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-30846

## Description 

Since Symfony 4.3 configuration root name should be passed via constructor instead of using "Symfony\Component\Config\Definition\Builder\TreeBuilder::root()" method.

More information:
* https://symfony.com/blog/new-in-symfony-4-2-important-deprecations#deprecated-tree-builders-without-root-nodes
* https://github.com/symfony/symfony/blob/master/UPGRADE-4.2.md#config

### DIFF
--- a/bundle/DependencyInjection/Configuration.php
+++ b/bundle/DependencyInjection/Configuration.php
@@ -42,8 +42,9 @@ class Configuration implements ConfigurationInterface
 
     public function getConfigTreeBuilder()
     {
-        $treeBuilder = new TreeBuilder();
-        $rootNode = $treeBuilder->root($this->rootNodeName);
+        $treeBuilder = new TreeBuilder($this->rootNodeName);
+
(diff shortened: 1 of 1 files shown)


bundle/DependencyInjection/Configuration.php
```diff
@@ -42,8 +42,9 @@ public function __construct($rootNodeName)
 
     public function getConfigTreeBuilder()
     {
-        $treeBuilder = new TreeBuilder();
-        $rootNode = $treeBuilder->root($this->rootNodeName);
+        $treeBuilder = new TreeBuilder($this->rootNodeName);
+
```
mikadamczyk:
```suggestion
```

PR #149, opened 2019-09-17
EZP-30951: Searching for Content with empty field will result on Error when using Solr

mateuszbieniek:
JIRA issue: https://jira.ez.no/browse/EZP-30951

This PR adds handling for Field Criterion where value `NULL` is passed for `EQ`.

### DIFF
--- a/lib/Query/Common/CriterionVisitor/Field/FieldIn.php
+++ b/lib/Query/Common/CriterionVisitor/Field/FieldIn.php
@@ -56,19 +56,27 @@ class FieldIn extends Field
             );
         }
 
-        $criterion->value = (array)$criterion->value;
+        if ($criterion->value === null) {
+            $criterion->value[] = null;
+        } else {
+            $criterion->value = (array)$criterion->value;
+        }
         $queries = array();
 
         foreach ($searchFields as $name => $fieldType) {
             foreach ($criterion->value as $value) {
-                $preparedValue = $this->escapeQuote(
-                    $this->toString(
-                        $this->mapSearchFieldValue($value, $fieldType)
-                    ),
-                    true
-                );
+                if ($value === null) {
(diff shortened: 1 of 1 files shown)


andrerom:
And does this even work on LegacySearch? Afaik IsEmpty Criterion is the missing feature for this.

mateuszbieniek:
@andrerom yeah, it works correctly in LegacySearch.

mateuszbieniek:
Side note: failing tests are not related to this PR.

mateuszbieniek:
@andrerom @alongosz I have a problem here which came out after I have written missing Integration Test - maybe you have any idea how to handle it:
In eZ Platform fields that have `null` value are not stored in Solr (so, `null` is basically an absence of field for document).

This part: https://github.com/ezsystems/ezplatform-solr-search-engine/pull/149/files#diff-01ed98326f3cd2df926d20e7b8386feeR69 doest exactly this: returns all documents that do not have field `X` in them, which will return not only Content with field `X` set to null but as well all other Contents that does not have `X` field in ContentType.

Therefore, currently, I don't see a way to fix it - or even if it is possible.

The best way would be to handle `null` with some default value, which would be stored in Solr and then there it is possible to check against it - but that would be big BC break - maybe it is a valid candidate for v3?

The second approach would be to somewhat retrieve a list of ContentTypes, check which ones have this field defined and then add an additional condition for the query (ContentType) so behaviour would be the same as for Legacy, but I have no idea how to achieve it without a significant performance hit.

A third possible approach is to leave it as it is, but with a note that it should only be used with Solar alongside `ContentTypeIdentifier`.

Forth one is to create a custom Criterion, that would merge Field and ContentTypeIdentifier Critetions into one.

WDYT?

mateuszbieniek:
Close in favor https://github.com/ezsystems/ezplatform-solr-search-engine/pull/157

PR #150, opened 2019-09-24
EZP-30296: Large API subtree delete leads to serious Solr index inconsistencies

kmadejski:
JIRA ticket: https://jira.ez.no/browse/EZP-30296

Bulk deleting using `deleteByQuery` has been introduced. I couldn't find particular information about Solr's bulk operation size limit, at least not in their documentation, but I think it's good to have some limit to prevent exceeding GET length limit for instance.

After a talk with @andrerom and @alongosz I decided to move further work on slots to a separated PR. Therefore, this one covers only bulk deleting.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": "~5.6|~7.0",
-        "ezsystems/ezpublish-kernel": "~6.7.10@dev|^6.13.6@dev|~7.3.5@dev|^7.4.3@dev",
+        "ezsystems/ezpublish-kernel": "~6.7.10@dev|^6.13.6@dev|~7.3.5@dev|~7.4.3@dev",
         "netgen/query-translator": "^1.0"
     },
     "require-dev": {
--- a/lib/Handler.php
+++ b/lib/Handler.php
@@ -45,6 +45,8 @@ use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
  */
 class Handler implements SearchHandlerInterface, Capable
 {
+    const SOLR_BULK_REMOVE_LIMIT = 1000;
+
     /**
      * Content locator gateway.
      *
@@ -340,9 +342,15 @@ class Handler implements SearchHandlerInterface, Capable
             $this->gateway->searchAllEndpoints($query)
         );
 
+        $contentDocumentIds = [];
+
         foreach ($searchResult->searchHits as $hit) {
-            $idPrefix = $this->mapper->generateContentDocumentId($hit->valueObject->id);
-            $this->gateway->deleteByQuery("_root_:{$idPrefix}*");
+            $contentDocumentIds[] = $this->mapper->generateContentDocumentId($hit->valueObject->id) . '*';
+        }
+
+        foreach (\array_chunk($contentDocumentIds, self::SOLR_BULK_REMOVE_LIMIT) as $ids) {
+            $query = '_root_:(' . implode(' OR ', $ids) . ')';
+            $this->gateway->deleteByQuery($query);
         }
     }
 
--- a/lib/Handler.php
+++ b/lib/Handler.php
@@ -47,8 +47,8 @@ class Handler implements SearchHandlerInterface, Capable
 {
     /* Solr's maxBooleanClauses config value is 1024 */
     const SOLR_BULK_REMOVE_LIMIT = 1000;
-    /* 32b max integer value due to Solr (JVM) limitations */
-    const SOLR_MAX_QUERY_LIMIT = PHP_INT_MAX >> 32;
+    /* 16b max unsigned integer value due to Solr (JVM) limitations */
+    const SOLR_MAX_QUERY_LIMIT = 65535;
     const DEFAULT_QUERY_LIMIT = 1000;
 
     /**
--- a/lib/Handler.php
+++ b/lib/Handler.php
@@ -45,7 +45,11 @@ use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
  */
 class Handler implements SearchHandlerInterface, Capable
 {
+    /* Solr's maxBooleanClauses config value is 1024 */
     const SOLR_BULK_REMOVE_LIMIT = 1000;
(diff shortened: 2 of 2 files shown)


lib/Resources/config/container/solr/slots.yml
```diff
@@ -15,7 +15,7 @@ parameters:
     ezpublish.search.solr.slot.move_user_group.class: eZ\Publish\Core\Search\Common\Slot\MoveUserGroup
     ezpublish.search.solr.slot.delete_user_group.class: eZ\Publish\Core\Search\Common\Slot\DeleteUserGroup
     ezpublish.search.solr.slot.copy_subtree.class: eZ\Publish\Core\Search\Common\Slot\CopySubtree
-    ezpublish.search.solr.slot.move_subtree.class: eZ\Publish\Core\Search\Common\Slot\MoveSubtree
+    ezpublish.search.solr.slot.move_subtree.class: EzSystems\EzPlatformSolrSearchEngine\Slot\MoveSubtree
```
andrerom: So I'm a bit torn here
- bulkIndexContent should be added to a SPI interface and handler here and others that can should  implement it. then we can update kernel's AbstractSubtree::indexSubtree() to take advantage.

If we don't do that, we'll need to overload all subtree slots here _(missing is hide and unhide at least from what I could see now in 6.13 branch)_.

andrerom:
@kmadejski Can you do a amend on commit message and force push to get Travis to run test with the right target branch?

kmadejski:
It seems that my PR is failing in the same places as the one done by @mateuszbieniek (https://github.com/ezsystems/ezplatform-solr-search-engine/pull/149). I'll try to fix these tests, as I would love to have Travis green before merge.

mateuszbieniek:
yes please :grin:

andrerom:
> It seems that my PR's tests are failing in the same places as the one done by @mateuszbieniek (#149). I'll try to fix these tests, as I would love to have Travis green before merge.

+1 _(missing backports from 1.6/1.7 maybe, which fails with test on kernel 7.5? if so maybe solution is to stop supporting 7.5 on 1.5 branch 🤷‍♂ and make it 1.x only)_

Regarding the PHP 5.6 failure on composer running out of memory, I think QA might have worked quite a bit on solving this other places cc @mnocon

mnocon:
I can't help you with the failiing tests, but I have a solution for memory issues 😉 

See https://github.com/ezsystems/ezplatform-solr-search-engine/pull/151

kmadejski:
I'm almost a 100% sure that there failing tests are related to a missing commit, the one from this PR: https://github.com/ezsystems/ezplatform-solr-search-engine/pull/133 to be more precise.
I've temporarily pushed it here to check whether it indeed solves the problem. Once confirmed, I'll revert this commit here and will push it directly to `1.5`.

kmadejski:
@andrerom last build (https://travis-ci.org/ezsystems/ezplatform-solr-search-engine/jobs/589992661) confirms my suspicion. How would you like to proceed? Personally, I would drop the possibility to install `1.5` for v2.x.

kmadejski:
Fixed issue found by @micszo in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/150/commits/4a99ddf65fadb13e652bc8dafd7faf58237e81ff.

composer.json
```diff
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": "~5.6|~7.0",
-        "ezsystems/ezpublish-kernel": "~6.7.10@dev|^6.13.6@dev|~7.3.5@dev|^7.4.3@dev",
+        "ezsystems/ezpublish-kernel": "~6.7.10@dev|^6.13.6@dev|~7.3.5@dev|~7.4.3@dev",
```
andrerom: I'm ok with this, so +1 once PR is green ;)

*TODO*: Drop this line on merge up.

lib/Handler.php
```diff
@@ -328,7 +334,7 @@ public function commit($flush = false)
      */
     protected function deleteAllItemsWithoutAdditionalLocation($locationId)
     {
-        $query = $this->prepareQuery();
+        $query = $this->prepareQuery(self::SOLR_MAX_QUERY_LIMIT);
```
andrerom: Ok I get why you wanted really high limit now..

I think unfortunately you'll have to change logic to iterate on this if you get back this amount of hits, or you might get total count which will tell you how many iterations you need to make right?

kmadejski: > I think unfortunately you'll have to change logic to iterate on this if you get back this amount of hits, or you might get total count which will tell you how many iterations you need to make right?

It was my first idea, but @adamwojs was really against this approach since it might be expensive, performance wise, to perform two same queries, first to count and second to load locations. WDYT @andrerom?

andrerom: But that said, that can be follow also. You definitely improve the logic here as previous limit was 1000...

So ignore this for now unless you feel up for it in this PR

andrerom: AS said above, this is already way better then what is there today. So I'm fine with this as is.

kmadejski: @andrerom cool, let's stay with what we have now. I will talk to @adamwojs once we both have time and try to figure out some better way to improve it perhaps.

kmadejski:
@micszo since it's green now, could you please test it again?

lserwatka:
@kmadejski you can merge it up but please remember about dropping line from composer.json

lserwatka:
@kmadejski make sure you merge to 1.7, 2.0 and master (3.0)

kmadejski:
@lserwatka done in https://github.com/ezsystems/ezplatform-solr-search-engine/commit/e66f9e884136bd7b2808065b657c82a9c5455b8a / https://github.com/ezsystems/ezplatform-solr-search-engine/commit/205a2565be8d4fafdeec43bc07573705b23705df / https://github.com/ezsystems/ezplatform-solr-search-engine/commit/f95f0f325a9f8a8aba0bb85e3ddb2ad483b32a93 / https://github.com/ezsystems/ezplatform-solr-search-engine/commit/2533029931946da47bbff2b37028556dec585c9a

PR #152, opened 2019-09-30
[composer] Set branch alias to 3.0.x-dev

adamwojs:
Updated branch-alias for dev-master to `3.0.x` in `composer.json` according to https://github.com/ezsystems/ezplatform-solr-search-engine/pull/136:

> Due to the major changes in Solr, support for 7.X will be part of the ezplatform-solr-search-engine 2.0 release which will be compatible with eZ Platform 2.5 release. Solr 6.X will be dropped in this release.
>
> Solr 7.X and Solr 8.X will be officialy supported in ezplatform-solr-search-engine 3.0 release which will be compatible with eZ Platform 3.X.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0.x-dev"
+            "dev-master": "3.0.x-dev"
         }
     }
 }


mnocon:
I'd change the value to 2.0.x-dev here: https://github.com/ezsystems/ezplatform-solr-search-engine/blob/2.0/composer.json#L37 and the  change it to 3.0.x-dev when merging upstream (that's how I've done it when setting dependencies for 3.0).

Also, you'll need a companion PR so that 3.0 (in solr) is used with eZ Platform 3.0: https://github.com/ezsystems/ezplatform/blob/master/composer.json#L30

PR #155, opened 2019-10-03
EZP-30985: Replaced deprecated DI short factory syntax

emodric:


alongosz:
Tested with eZ Platform EE Demo (reindexing on Solr).
CI failure is unrelated.

PR #157, opened 2019-10-11
EZP-30951: Added missing support for searching (Not)Empty Field values

mateuszbieniek:
| Question           | Answer
| **JIRA issue**     | [EZP-30951](https://jira.ez.no/browse/EZP-30951)
| **Requires** | ezsystems/ezpublish-kernel#2821
| **Bug/Improvement**| yes
| **New feature**    | yes
| **Target version** | `1.7` for eZ Platform `v2.5.x` LTS
| **Doc needed** | yes

This PR adds Mapper, so a new index for not required fields will be added to Solr: `is_empty`.
`is_empty` (true/false) determines if Field is considered empty as per Feild Definition.

The Developer then can search (not) empty Fields with `IsFieldEmpty` Criterion.

This PR requires these kernel changes: https://github.com/ezsystems/ezpublish-kernel/pull/2821


alongosz:
@mateuszbieniek could you provide, either in this PR description or in the kernel one, a sample PHP code, so it's easier for both the Doc Team and the others to understand how they can use it?

The scope is just a few lines, starting from building Criteria and ending with calling e.g. `findContent`.

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentNullFields.php
```diff
@@ -0,0 +1,91 @@
...
+        foreach ($content->fields as $field) {
+            if ($field->languageCode !== $languageCode) {
+                continue;
+            }
+
+            foreach ($contentType->fieldDefinitions as $fieldDefinition) {
+                if ($fieldDefinition->id !== $field->fieldDefinitionId) {
+                    continue;
+                }
+                if ($fieldDefinition->isRequired) {
```
kmadejski: This condition is cheaper than:
```
if ($fieldDefinition->id !== $field->fieldDefinitionId) {
    continue;
}
```
please swap them.

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentNullFields.php
```diff
@@ -0,0 +1,91 @@
...
+                    continue;
+                }
+
+                $fields[] = new Field(
+                    $name = $this->fieldNameGenerator->getName(
+                        self::IS_NULL_NAME,
+                        $fieldDefinition->identifier
+                    ),
+                    (null == $field->value->data),
```
kmadejski:
```suggestion
                    (null === $field->value->data),
```

lib/Query/Common/CriterionVisitor/Field/FieldEmpty.php
```diff
@@ -0,0 +1,96 @@
...
+
+        if (empty($searchFields)) {
+            throw new InvalidArgumentException(
+                '$criterion->target',
+                "No searchable fields found for the given criterion target '{$criterion->target}'."
+            );
+        }
+
+        $criterion->value = (array)$criterion->value;
+        $queries = array();
```
kmadejski:
```suggestion
        $queries = [];
```

lib/Query/Common/CriterionVisitor/Field/FieldIn.php
```diff
@@ -56,19 +78,33 @@ public function visit(Criterion $criterion, CriterionVisitor $subVisitor = null)
             );
         }
 
-        $criterion->value = (array)$criterion->value;
+        if ($criterion->value === null) {
```
kmadejski: It doesn't really change anything, but what do you think about:
```
$criterion->value = is_null($criterion->value) ? [null] : (array)$criterion->value;
```

mateuszbieniek:
@alongosz sure thing! Ill do it on monday :)

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentEmptyFields.php
```diff
@@ -0,0 +1,93 @@
...
+                }
+
+                /** @var \eZ\Publish\Core\Persistence\FieldType $fieldType */
+                $fieldType = $this->fieldTypeRegistry->getFieldType($fieldDefinition->fieldType);
+                $fields[] = new Field(
+                    $name = $this->fieldNameGenerator->getName(
+                        self::IS_EMPTY_NAME,
+                        $fieldDefinition->identifier
+                    ),
+                    ($fieldType->getEmptyValue() == $field->value),
```
pspanja: This should use `$fieldType->isEmptyValue($field->value)` instead.

pspanja:
What is the use case for searching that field value is equal to `null`?

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentEmptyFields.php
mateuszbieniek: Am I missing something? `\eZ\Publish\Core\Persistence\FieldType` has no such method.
https://github.com/ezsystems/ezpublish-kernel/blob/7.5/eZ/Publish/Core/Persistence/FieldType.php

mateuszbieniek:
@pspanja to match Legacy behaviour and to allow search for fields that are `null` - for example, `ezstring` can be empty, but not null. Two different things.

lib/Query/Common/CriterionVisitor/Field/FieldIn.php
mateuszbieniek: Decrease of readability exchanged only for a lesser number of lines IMHO - not a fan.

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentEmptyFields.php
pspanja: It needs to be added there :)

pspanja:
@mateuszbieniek I still don't see how it would be useful. For example, what concrete usecase would it solve?

mateuszbieniek:
@pspanja one example could be middle name: empty when not specifed, null when person has none.

pspanja:
That's actually not properly supported across field types, `null` is an alias for an empty value, which is an empty string in case of the TextLine.

mateuszbieniek:
@pspanja what value is empty is defined per FieldType. I don't really see a downside of the possibility to distinguish empty strings from nulls (in `ezstring` example). Please mind that you can develop your own FieldTypes with you can find `null` searching even more usable - adding this is a low hanging fruit and a useful one.

Can you please explain why are you against it?

pspanja:
@mateuszbieniek I've already said why, "no value" is not properly supported by the FieldType implementation. In case of TextLine, when you load the Content with such field its value will be an empty string, not `null`. That means with this you are actually formalizing something that is a detail of the Legacy Storage. Proper support might include the ability to do something like `$fieldType->hasValue($field)`.

You probably can account for that and make it consistent across FieldTypes. For me it's not a big deal, I just wanted to understand what is the intention behind it.

mateuszbieniek:
@pspanja maybe I can use a different example: I have a Custome FieldType that I use to display image. For me, empty value there is "blank.jpg" - not null, not empty string - just some placeholder "blank" image. Also, I want to be able to set this field to null - no image at all. That's the use case - empty value is defined per FieldType and it does not have to be null.

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentEmptyFields.php
alongosz: @mateuszbieniek you probably need here instances of `\eZ\Publish\SPI\FieldType\FieldType` instead of `\eZ\Publish\SPI\Persistence\FieldType`. That former one has this method. 

It also exists on `\eZ\Publish\API\Repository\FieldType`. The easiest way to obtain it is by using `\eZ\Publish\API\Repository\FieldTypeService` injectable by `ezpublish.api.service.field_type`.

mateuszbieniek: Thanks @alongosz !

pspanja: You should not use API FieldType in the Persistence implementation, that would be breaking the layers and why there is SPI/Persistence/FieldType version of the interface in the first place.

You can implement it like this: https://github.com/netgen/ezplatform-search-extra/blob/master/lib/Core/Persistence/FieldType.php#L15-L20

mateuszbieniek:
Rebased to `1.7`

mateuszbieniek:
After investigating `Field` Criterion and its behavior in `LegacySearchEngine` I noticed that basically all it is doing is castin `null` to an empty string. Therefore I decided to drop `is_null` index for Solr.
So in short: @pspanja, you were right :)

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentEmptyFields.php
```diff
@@ -0,0 +1,105 @@
+<?php
+
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
```
adamwojs: Could you please correct the file header ? 

```
/**
 * @copyright Copyright (C) eZ Systems AS. All rights reserved.
 * @license For full copyright and license information view LICENSE file distributed with this source code.
 */
declare(strict_types=1);

```

lib/Query/Common/CriterionVisitor/Field/FieldEmpty.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+/**
+ * This file is part of the eZ Platform Solr Search Engine package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
```
adamwojs: The file header is incorrect

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentEmptyFields.php
andrerom: @alongosz I'm a bit concerned by hardcoding `$fieldType->getEmptyValue() == $field->value`

If your reasoning for not introducing this on 2.5 is BC, then lets add a new interface for it and detect that here making it opt in*, optionally we can clean that up in either directly in 3.x or over time towards 4.x.

This is how Symfony deal with this kind of BC topics, they don't block minor features on LTS branches needed to solve bugs due to BC reasons, they just make sure they don't break stuff while still making sure bugs are solved in a clean way.

* like
```php

                /** @var \eZ\Publish\Core\Persistence\FieldType $fieldType */
                $fieldType = $this->fieldTypeRegistry->getFieldType($fieldDefinition->fieldType);
                // Check if we have the method to check if field is empty, skip if missing (older kernel version)
                if (!$fieldType instanceof Persistence\IsEmptyFieldType) {
                    continue;
                }
```

alongosz: @andrerom we could do that, but we should introduce strict return type to avoid creating technical debt. So if we add that the way I described it, it will cause a break in @pspanja's code regardless of if we create interface or not, because we need to add it to Core Field Type.

Alternatively we could go for impl. w/o return type for `7.5` and add it instantly on `8.0`, but this still would break Petar's code. Though would be BC-safe, because 8.0 is unstable by definition.

_Of course if I haven't seen that code, I wouldn't even consider questioning interface approach_

andrerom: > So if we add that the way I described it, it will cause a break in @pspanja's code regardless of if we create interface or not, because we need to add it to Core Field Type.

true
What's your take here @pspanja? One one side you seem to want this added, on the other side it will break your code here right?

pspanja: I'm not sure I completely follow, this does not need to be added to the Core FieldType, Persistence already has it's own FieldType interface where this can be added: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/SPI/Persistence/FieldType.php

From my side I think I will be able adapt, so go ahead and do it :)

pspanja: BTW, we have FieldType API integration testing framework, this can be updated so that each FieldType gets tested with `IsEmptyField` criterion: https://github.com/ezsystems/ezpublish-kernel/tree/master/eZ/Publish/API/Repository/Tests/FieldType

alongosz: > BTW, we have FieldType API integration testing framework, this can be updated so that each FieldType gets tested with `IsEmptyField` criterion: https://github.com/ezsystems/ezpublish-kernel/tree/master/eZ/Publish/API/Repository/Tests/FieldType

good point // cc @mateuszbieniek 

> I'm not sure I completely follow, this does not need to be added to the Core FieldType, Persistence already has it's own FieldType interface where this can be added: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/SPI/Persistence/FieldType.php

Which is implemented by `\eZ\Publish\Core\Persistence\FieldType` which is extended by `\Netgen\EzPlatformSearchExtra\Core\Persistence\FieldType`, so adding 
``` php
    public function isEmptyValue(FieldValue $fieldValue): bool
    {
        return $this->internalFieldType->isEmptyValue(
            $this->internalFieldType->fromPersistenceValue($fieldValue)
        );
    }
```

would break that because of the return type.

> From my side I think I will be able adapt, so go ahead and do it :)

If you could, then it would simplify things :)

As for SPI interface, cannot do it for eZ Platform 2.5.x, we can do it for 3.0. Therefore suggestion of another interface which would, I assume, disappear in 3.0.

pspanja: Ok, got it. I think that would be OK, I can probably work around it with `__call`, in that case it won't be a big deal at all.

lib/Query/Common/CriterionVisitor/Field/FieldEmpty.php
```diff
@@ -0,0 +1,97 @@
...
+     * @return string
+     */
+    public function visit(Criterion $criterion, CriterionVisitor $subVisitor = null)
+    {
+        $searchFields = $this->getSearchFields($criterion);
+
+        if (empty($searchFields)) {
+            throw new InvalidArgumentException(
+                '$criterion->target',
+                "No searchable fields found for the given criterion target '{$criterion->target}'."
```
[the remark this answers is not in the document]
mateuszbieniek: This actually mimics behavior form `FieldIn` Criterion Visitor to keep consistent:
https://github.com/ezsystems/ezplatform-solr-search-engine/blob/1.7/lib/Query/Common/CriterionVisitor/Field/FieldIn.php#L53

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentEmptyFields.php
```diff
@@ -0,0 +1,108 @@
...
+        FieldTypeRegistry $fieldTypeRegistry
+    ) {
+        $this->contentTypeHandler = $contentTypeHandler;
+        $this->fieldNameGenerator = $fieldNameGenerator;
+        $this->fieldTypeRegistry = $fieldTypeRegistry;
+    }
+
+    /**
+     * @param \eZ\Publish\SPI\Persistence\Content $content
+     * @param $languageCode
```
alongosz: missing type-hint

lib/FieldMapper/ContentTranslationFieldMapper/ContentDocumentEmptyFields.php
```diff
@@ -0,0 +1,108 @@
...
+
+            foreach ($contentType->fieldDefinitions as $fieldDefinition) {
+                if ($fieldDefinition->isRequired) {
+                    continue;
+                }
+                if ($fieldDefinition->id !== $field->fieldDefinitionId) {
+                    continue;
+                }
+
+                /** @var \eZ\Publish\Core\Persistence\IsEmptyFieldType $fieldType */
```
alongosz: Needs alignment after the most recent changes to the kernel PR

lserwatka:
You can merge it up

mateuszbieniek:
@lserwatka done.

PR #158, opened 2019-10-16
Updated README.md

adamwojs:
> JIRA: -

### Description 

Updated information in README about supported versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,21 +6,23 @@
 [![License](https://img.shields.io/github/license/ezsystems/ezplatform-solr-search-engine.svg?style=flat-square)](LICENSE)
 
 Solr Search Engine Bundle for use with:
-- v1.6+: eZ Platform 2.5LTS+ *(bundled out of the box)* with Solr 6.x _(recommended: 6.6 which is an LTS)_
+- v3.0: eZ Platform 3.X with Solr 7.x _(recommended: 7.7. which is an LTS)_ 
+- v2.0: eZ Platform 2.5LTS with Solr 7.x _(recommended: 7.7. which is an LTS)_
+- v1.7: eZ Platform 2.5LTS+ *(bundled out of the box)* with Solr 6.x _(recommended: 6.6 which is an LTS)_
 - v1.5: eZ Platform 1.7LTS & 1.13LTS *(bundled out of the box)* with Solr 6.x or 4.10.4 _(recommended: 6.6 which is an LTS, and certain features only work on Solr 6)_
 - v1.0.x: eZ Publish Platform Enterprise 5.4.5+ *(optional, not as feature rich but helpful for scaling filtering queries)* with Solr 4.10.4
 
 #####  Overview of features
 
-| Feature                       | Solr Search Engine        | Legacy Search Engine _(SQL)_ |
-|:------------------------------|:-------------------------:|:----------------------------:|
-| Filtering                     | Yes                       | Yes, but limited*            |
-| Query _(filter with scoring)_ | Yes                       | Only filters, no scoring     |
-| Fulltext                      | Yes, incl advance features| Partly**                     |
-| Faceting                      | [Partly as of v1.4][1]    | No                           |
-| Spellchecking                 | Planned _(TBD when)_      | No                           |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,21 +6,23 @@
 [![License](https://img.shields.io/github/license/ezsystems/ezplatform-solr-search-engine.svg?style=flat-square)](LICENSE)
 
 Solr Search Engine Bundle for use with:
-- v1.6+: eZ Platform 2.5LTS+ *(bundled out of the box)* with Solr 6.x _(recommended: 6.6 which is an LTS)_
+- v3.0: eZ Platform 3.X with Solr 7.x _(recommended: 7.7. which is an LTS)_ 
+- v2.0: eZ Platform 2.5LTS with Solr 7.x _(recommended: 7.7. which is an LTS)_
+- v1.7: eZ Platform 2.5LTS+ *(bundled out of the box)* with Solr 6.x _(recommended: 6.6 which is an LTS)_
```
andrerom: We can drop this once 2.0.0 is tagged, so we don't need both here at the same time actually, or?

README.md
```diff
@@ -6,21 +6,23 @@
...
-| Index time boosting           | [Yes, as of v1.4][4]      | No                           |
+| Feature                       | Solr Search Engine          | Legacy Search Engine _(SQL)_ |
+|:------------------------------|:---------------------------:|:----------------------------:|
+| Filtering                     | Yes                         | Yes, but limited*            |
+| Query _(filter with scoring)_ | Yes                         | Only filters, no scoring     |
+| Fulltext                      | Yes, incl advance features  | Partly**                     |
+| Faceting                      | [Partly as of v1.4][1]      | No                           |
+| Spellchecking                 | Planned _(TBD when)_        | No                           |
+| Highlighting                  | Planned _(TBD when)_        | No                           |
+| Index time boosting           | [Yes, in >= v1.4,< v2.0][4] | No                           |
```
andrerom: ?? Should this be expanded in not 4?

adamwojs: Index time boosting has been removed in Solr 7.7 completely:

> Index-time boosts have been removed from Lucene, and are no longer available from Solr. If any boosts are provided, they will be ignored by the indexing chain. As a replacement, index-time scoring factors should be indexed in a separate field and combined with the query score using a function query. See the section Function Queries for more information.

Ref. https://lucene.apache.org/solr/guide/7_2/major-changes-in-solr-7.html 

That was one of the reasons why we decided to go with major release of ezplatform-solr-search-engine.

I'm working on query time boosting (https://github.com/ezsystems/ezplatform-solr-search-engine/compare/ezp_30785) so the replacement should be available in the incoming v2.1.

README.md
adamwojs: IMHO there is no reason to start a new project with Solr 6.X and I would be more than happy to drop 1.7 support for eZ Platform 2.5 but switching from ezplatform-solr-search-engine from 1.X to 2.X also implies Solr upgrade so maybe let's give some time for already running projects to migrate 😉

README.md
andrerom: Clear, can you expand note 4 to explain this or add new note for this?

README.md
andrerom: OK 👍

README.md
adamwojs: Fixed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/158/commits/2c88d8ac59dc3039eb5f8b22788658cb3f4ad5b7

README.md
```diff
@@ -29,6 +31,9 @@ _** Does not include full set of full text features includes: Scoring/Ranking _(
 by default, for location search available as of [v1.3][3] with Solr 6.x)_, and as of [Solr Search Engine v1.5][5]
 supports advanced full text capabilities such as `word "phrase" (group) +mandatory -prohibited AND && OR || NOT !`_
 
+_*** Index-time boosts have been removed from Lucene, and are no longer available from Solr 7.
```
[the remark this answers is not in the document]
andrerom:
```suggestion
_*** Index-time boosts have been removed from Lucene and are no longer available from Solr 7, we plan to introduce query time boosting instead in future 2.x and 3.x release.
```

PR #159, opened 2019-10-25
EZP-30759: Creating content with terminal escape codes chars inside Searchable field will return 500 error when using Solr

kmadejski:
| Question           | Answer
| **JIRA issue**     | [EZP-30759](https://jira.ez.no/browse/EZP-30759)
| **Bug**| yes
| **New feature**    | no
| **Target version** | `1.5` and higher

This PR fixes the issue described in the mentioned JIRA ticket. The simple solution is to filter out all the terminal escape characters but just for Solr as Legacy storage handles it without any issue. 

I'll add type hints on merge up, as `1.5` can be used with eZ Platform v1.x on PHP 5.x.

I'm about to create an issue in Solr's JIRA tracker, but as for now, I'm waiting for their devs to confirm that from their perspective this can be considered as a bug. I'll update the PR and add a comment in the code once the issue is created.

### DIFF
--- a/lib/FieldMapper/ContentFieldMapper/BlockDocumentsBaseContentFields.php
+++ b/lib/FieldMapper/ContentFieldMapper/BlockDocumentsBaseContentFields.php
@@ -43,21 +43,29 @@ class BlockDocumentsBaseContentFields extends ContentFieldMapper
     protected $sectionHandler;
 
     /**
+     * @var string
+     */
+    private $invalidCharactersPattern;
+
+    /**
      * @param \eZ\Publish\SPI\Persistence\Content\Location\Handler $locationHandler
      * @param \eZ\Publish\SPI\Persistence\Content\Type\Handler $contentTypeHandler
      * @param \eZ\Publish\SPI\Persistence\Content\ObjectState\Handler $objectStateHandler
      * @param \eZ\Publish\SPI\Persistence\Content\Section\Handler $sectionHandler
+     * @param string $invalidCharactersPattern
      */
     public function __construct(
         LocationHandler $locationHandler,
         ContentTypeHandler $contentTypeHandler,
         ObjectStateHandler $objectStateHandler,
-        SectionHandler $sectionHandler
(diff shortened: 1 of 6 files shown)


lib/Resources/config/container/solr/services.yml
```diff
@@ -10,6 +10,7 @@ parameters:
     ezpublish.search.solr.field_mapper.location.class: EzSystems\EzPlatformSolrSearchEngine\FieldMapper\LocationFieldMapper\Aggregate
     ezpublish.search.solr.field_mapper.boost_factor_provider.class: EzSystems\EzPlatformSolrSearchEngine\FieldMapper\BoostFactorProvider
     ezpublish.search.solr.field_mapper.boost_factor_provider.map: []
+    ez_search_engine_solr.invalid_characters_pattern: "#\\x1b[[][^A-Za-z]*[A-Za-z]#"
```
andrerom: Comment with link to Solr issue here so we can track it.

lib/FieldMapper/ContentFieldMapper/BlockDocumentsBaseContentFields.php
```diff
@@ -42,22 +42,30 @@ class BlockDocumentsBaseContentFields extends ContentFieldMapper
      */
     protected $sectionHandler;
 
+    /**
+     * @var string
```
andrerom: Maybe something like this (here and on the others):
```suggestion
     * @var string Pattern with special characters that should be stripped from field value due to Solr not accepting them.
```

gggeek: 'regexp pattern' ?

gggeek:
-1. We should quote special chars instead of stripping them out. Esp. if this is done pervasively. Do you prohibit single quotes chars from ever being stored in varchar fields in databases?

gggeek:
ps: I agree about filtering out a very limited set of characters, but not fe. '['

PR #169, opened 2020-02-24
EZP-31080: Created missing search Criterions

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-31080

Implemented visitors for new criterions:
* ObjectStateIdentifier
* SectionIdentifier

I wanted to add more visitors but it isn't that easy. All user based criterions will require creating new document type so it's out of scope for now.

### DIFF
--- a/lib/FieldMapper/ContentFieldMapper/BlockDocumentsBaseContentFields.php
+++ b/lib/FieldMapper/ContentFieldMapper/BlockDocumentsBaseContentFields.php
@@ -177,6 +177,11 @@ class BlockDocumentsBaseContentFields extends ContentFieldMapper
                 $this->getObjectStateIds($contentInfo->id),
                 new FieldType\MultipleIdentifierField()
             ),
+            new Field(
+                'content_object_state_identifiers',
+                $this->getObjectStateIdentifiers($contentInfo->id),
+                new FieldType\MultipleStringField()
+            ),
         ];
     }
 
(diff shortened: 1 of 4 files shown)


lib/Query/Common/CriterionVisitor/ObjectStateIdentifierIn.php
```diff
@@ -0,0 +1,42 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
```
adamwojs: The license header is in the wrong place (also in all other new files).

lib/FieldMapper/ContentFieldMapper/BlockDocumentsBaseContentFields.php
```diff
@@ -201,6 +206,27 @@ protected function getObjectStateIds($contentId)
...
+        foreach ($this->objectStateHandler->loadAllGroups() as $objectStateGroup) {
+            $identifiers[] = sprintf(
+                '%s:%s',
+                $objectStateGroup->identifier,
+                $this->objectStateHandler->getContentState(
+                    $contentId,
+                    $objectStateGroup->id
+                )->identifier
+            );
+        }
```
alongosz: As long as this Criterion is modeled properly, you're not going to need this, see https://github.com/ezsystems/ezpublish-kernel/pull/2952#pullrequestreview-371896366.

PR #173, opened 2020-03-05
EZP-31420: Deleted translation from Solr when deleted from Version

mateuszdebinski:
| Question           | Answer
| **JIRA issue**     | [EZP-31420](https://jira.ez.no/browse/EZP-31420)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | `1,7`
| **BC breaks**      | no
| **Tests pass**     | [yes](https://travis-ci.org/github/ezsystems/ezpublish-kernel/builds/675704022)
| **Doc needed**     | no

## Description 
Depends on: https://github.com/ezsystems/ezpublish-kernel/pull/3013

Removing a content translation does not remove the associated SolR documents

**TODO**:
- [X] Implement feature / fix a bug.
- [x] Ask for Code Review.

### DIFF
--- a/lib/Handler.php
+++ b/lib/Handler.php
@@ -478,11 +478,8 @@ class Handler implements SearchHandlerInterface, Capable, ContentTranslationHand
 
     /**
      * Deletes a translation content object from the index.
-     *
-     * @param int $contentId
-     * @param string $languageCode
      */
-    public function deleteTranslation($contentId, $languageCode)
+    public function deleteTranslation(int $contentId, string $languageCode): void
     {
         $languageCode = preg_replace('([^A-Za-z0-9/]+)', '', $languageCode);
         $idPrefix = $this->mapper->generateContentDocumentId($contentId, $languageCode);
--- a/lib/Handler.php
+++ b/lib/Handler.php
@@ -14,6 +14,7 @@ use eZ\Publish\API\Repository\SearchService;
 use eZ\Publish\SPI\Persistence\Content;
 use eZ\Publish\SPI\Persistence\Content\Location;
 use eZ\Publish\SPI\Persistence\Content\Handler as ContentHandler;
+use eZ\Publish\SPI\Search\ContentTranslationHandler;
 use eZ\Publish\SPI\Search\Handler as SearchHandlerInterface;
 use eZ\Publish\SPI\Search\Capable;
 use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
@@ -43,7 +44,7 @@ use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
  * content objects based on criteria, which could not be converted in to
  * database statements.
  */
-class Handler implements SearchHandlerInterface, Capable
+class Handler implements SearchHandlerInterface, Capable, ContentTranslationHandler
 {
     /* Solr's maxBooleanClauses config value is 1024 */
     const SOLR_BULK_REMOVE_LIMIT = 1000;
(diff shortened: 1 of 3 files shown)


tests/lib/Slot/RemoveTranslationTest.php
```diff
@@ -76,6 +76,13 @@ public function testReceive()
             ->willReturn($content)
         ;
 
+        $this
+            ->getSearchHandlerMock()
+            ->expects($this->once())
+            ->method('deleteContent')
+            ->with(2, 1)
+        ;
```
konradoboza: I think that this assertion should be placed before `load` and `indexContent` as deleting content appears before indexing.

mateuszdebinski: yes, you have right, corrected

lib/Slot/RemoveTranslation.php
```diff
@@ -34,6 +34,11 @@ public function receive(Signal $signal)
             return;
         }
 
+        $this->searchHandler->deleteContent(
+            $contentInfo->id,
+            $contentInfo->currentVersionNo
+        );
```
andrerom: Doesn't this risk index being w/o the content before indexing is done? (race condition)
So would be better if we expose a way to delete specific translation instead.

alongosz: > Doesn't this risk index being w/o the content before indexing is done? (race condition)
> So would be better if we expose a way to delete specific translation instead.

I fully agree. Moreover the fact that indexing will be done after that operation, though occurs, it's not a guarantee due to async nature of Signals. Might occur before, might not occur.

andrerom:
Oh, we need to handle DeleteTranslation as well right? As RemoveTranslation seems to be the deprecated one if I read @alongosz's master PR right.

alongosz:
Yes, RemoveTranslationSignal is emitted for BC. You could actually handle both RemoveTranslationSignal and DeleteTranslationSignal in one single DeleteTranslationSlot. Don't remember how it's handled on upper branch with Events though.

lib/Handler.php
```diff
@@ -474,4 +475,19 @@ public function supports($capabilityFlag)
...
+
+    /**
+     * Deletes a translation content object from the index.
+     *
+     * @param int $contentId
+     * @param string $languageCode
+     */
+    public function deleteTranslation($contentId, $languageCode)
+    {
+        $languageCode = preg_replace('([^A-Za-z0-9/]+)', '', $languageCode);
```
alongosz, line 487: what happened here?

mateuszdebinski: Convert to a proper search engine representation. used from this: https://github.com/ezsystems/ezpublish-kernel/blob/01cb7b9e9aae7727b63a9558988e79c33d2d1fe6/eZ/Publish/Core/Search/Common/FieldValueMapper/IdentifierMapper.php#L51

alongosz: > Convert to a proper search engine representation. used from this: https://github.com/ezsystems/ezpublish-kernel/blob/01cb7b9e9aae7727b63a9558988e79c33d2d1fe6/eZ/Publish/Core/Search/Common/FieldValueMapper/IdentifierMapper.php#L51

So why repeating it instead of re-using it? Via DI if available.

mateuszdebinski:
> +1, but as I said in [ezsystems/ezpublish-kernel#3013 (review)](https://github.com/ezsystems/ezpublish-kernel/pull/3013#pullrequestreview-401734722) I need to see a merge up plan, probably for 3.0 branch here (but including merge as-is to 2.0 due to gitflow).

Solr Slot moved to kernel event, so any more changes in 3.0

lserwatka:
You can merge it up.

PR #179, opened 2020-03-27
Specify port in solr_create_core for the tests

SerheyDolgushev:
I was running tests for https://github.com/ezsystems/ezplatform-xmltext-fieldtype on my local, where I have multiple Solr instances. After executing `./vendor/ezsystems/ezplatform-solr-search-engine/bin/.travis/init_solr.sh` cores were created in wrong Solr instance.
This PR fixes it.

### DIFF
--- a/bin/.travis/init_solr.sh
+++ b/bin/.travis/init_solr.sh
@@ -161,7 +161,7 @@ solr_create_core() {
     core_name=$1
     config_dir=$2
 
-    ./${SOLR_INSTALL_DIR}/bin/solr create_core -c ${core_name} -d ${config_dir} || exit_on_error "Can't create core"
+    ./${SOLR_INSTALL_DIR}/bin/solr create_core -p ${SOLR_PORT} -c ${core_name} -d ${config_dir} || exit_on_error "Can't create core"
 }
 
 solr_cloud_configure_nodes() {


alongosz:
> I was running tests for https://github.com/ezsystems/ezplatform-xmltext-fieldtype on my local, where I have multiple Solr instances. After executing `./vendor/ezsystems/ezplatform-solr-search-engine/bin/.travis/init_solr.sh` cores were created in wrong Solr instance.
> This PR fixes it.

As the name of the script suggests, it's meant for Travis, not you :) We don't provide support for it.
That said, maybe it should be executed like that anyway. POV ping @adamwojs.

SerheyDolgushev:
Yep, I understand this script is for Travis usage only. But I had no better idea how to try to reproduce the failing tests on my local,  except mimic the Travis env.
Personally, I think that this PR makes this script a bit more specific and clear. But I`m also very curious to hear what @adamwojs  will say.

SerheyDolgushev:
@emodric I`m almost sure this PR will not fix the failing tests but tagging you here just in case.

adamwojs:
@SerheyDolgushev +1 but please make sure it's works on default port if `SOLR_PORT` variable is not specified.

PR #187, opened 2020-06-01
Run integration tests on Solr 7.7.3 and 8.5.2

adamwojs:
> JIRA: - 

## Description 

Enabled Solr testing on 7.7.3 (last 7.X release) and 8.5.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,13 +13,22 @@ matrix:
         - php: 7.3
           env: TEST_CONFIG="phpunit.xml"
         - php: 7.3
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="7.7.2"  CORES_SETUP="dedicated"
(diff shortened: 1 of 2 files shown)


andrerom:
@adamwojs / @lserwatka Can we backport Solr 8.x support to 2.x branch too? 

Why (1): Unless Platform.sh adds support for Solr 7 we are actually in a bit strange situation with 2.5 _(and until this PR also for 3.x)_, as they don't support Solr 7 on "PE" Dedicated Cluster setup _(which most of our customers use)_.

Both me and @SylvainGuittard is trying to get them to add Solr 7, and I'll hear from them on thursday. But (2) as 2.x and 3.x branch are somewhat similar it might be good idea anyway to keep them in sync on things like this.

SylvainGuittard:
>Can we backport Solr 8.x support to 2.x branch too?
I already asked @adamwojs yesterday. It does not seem too complicated according to Adam. 
It will solve the current issue we have on grid right now "in a faster way".

adamwojs:
@andrerom / @SylvainGuittard We will do the backport :wink:

PR #208, opened 2021-02-02
Refactored ResultExtractor::extract method

adamwojs:
> JIRA: -

### Description 

Refactored `\EzSystems\EzPlatformSolrSearchEngine\ResultExtractor::extract` method by splitting it into several smaller methods. 

The main reason behind this PR is to simplify result extractor decoration in order to  to to be able to enrich search hits with custom metadata e.g. https://lucene.apache.org/solr/guide/7_3/spatial-search.html#SpatialSearch-geodist 

### Further refactoring

Next step in refactoring for 4.x is to extract SearchHit mapping logic  to dedicated ResutiHitExtractor service.

### DIFF
--- a/lib/ResultExtractor.php
+++ b/lib/ResultExtractor.php
@@ -16,6 +16,7 @@ use EzSystems\EzPlatformSolrSearchEngine\Query\FacetFieldVisitor;
 use eZ\Publish\API\Repository\Values\Content\Search\SearchResult;
 use eZ\Publish\API\Repository\Values\Content\Search\SearchHit;
 use EzSystems\EzPlatformSolrSearchEngine\ResultExtractor\AggregationResultExtractor;
+use stdClass;
 
 /**
  * Abstract implementation of Search Extractor, which extracts search result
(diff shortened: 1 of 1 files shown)


lib/ResultExtractor.php
```diff
@@ -62,64 +63,27 @@ public function extract($data, array $facetBuilders = [], array $aggregations =
...
+    /**
+     * Extracts value object from $hit returned by Solr backend.
+     *
+     * Needs to be implemented by the concrete ResultExtractor.
+     *
+     * @param mixed $hit
+     *
+     * @return \eZ\Publish\API\Repository\Values\ValueObject
+     */
+    abstract public function extractHit($hit);
```
Steveb-p, line 85: Can this be strictly typed?

```suggestion
    abstract public function extractHit($hit): ValueObject;
```

lib/ResultExtractor.php
```diff
@@ -151,13 +115,78 @@ protected function getIndexIdentifier($hit)
...
+
+            foreach ($data->facet_counts as $facetCounts) {
+                foreach ($facetCounts as $field => $facet) {
+                    if (empty($facetBuildersById[$field])) {
+                        @trigger_error(
+                            'Not setting id of field using FacetFieldVisitor::visitBuilder will not be supported in 2.0'
+                            . ', as it makes it impossible to exactly identify which facets belongs to which builder.'
+                            . "\nMake sure to adapt your visitor for the following field: ${field}"
+                            . "\nExample: 'facet.field' => \"{!ex=dt key=\${id}}${field}\",",
+                            E_USER_DEPRECATED);
```
Steveb-p, line 166: Are we against using https://github.com/symfony/deprecation-contracts here?

lib/ResultExtractor.php
```diff
@@ -151,13 +115,78 @@ protected function getIndexIdentifier($hit)
...
+                    }
+
+                    $facets[] = $this->facetBuilderVisitor->mapField(
+                        $field,
+                        (array)$facet,
+                        isset($facetBuildersById[$field]) ? $facetBuildersById[$field] : null
```
Steveb-p, line 172:
```suggestion
                        $facetBuildersById[$field] ?? null
```

lib/ResultExtractor.php
adamwojs: This would be BC break. Technically this class as is not part of the API/SPI so shouldn't be extended but reality is that in many cases there is no other choice.

alongosz: > This would be BC break. Technically this class as is not part of the API/SPI so shouldn't be extended but reality is that in many cases there is no other choice.

This is not supported.

lib/ResultExtractor.php
alongosz: > What's our approach for using https://github.com/symfony/deprecation-contracts for cases like this?

It's fairly fresh. We can use it, but we should all be on the same page with that to avoid creating debt.

adamwojs: Well, this code fragment is much older then symfony/deprecation-contracts. But I wouldn't spend time on something that we are going to remove very soon :wink:

lib/ResultExtractor.php
```diff
@@ -151,13 +115,78 @@ protected function getIndexIdentifier($hit)
...
+            $facetBuildersById = [];
+            foreach ($facetBuilders as $facetBuilder) {
+                $facetBuildersById[spl_object_hash($facetBuilder)] = $facetBuilder;
+            }
+
+            foreach ($data->facet_counts as $facetCounts) {
+                foreach ($facetCounts as $field => $facet) {
+                    if (empty($facetBuildersById[$field])) {
+                        @trigger_error(
+                            'Not setting id of field using FacetFieldVisitor::visitBuilder will not be supported in 2.0'
```
alongosz, line 162: Well... I think we're too late here.

alongosz (changes requested):

lib/ResultExtractor.php
adamwojs: Fixed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/208/commits/ecd3d65514f58e2be7863b56fb13f3bc1025884d

lib/ResultExtractor.php
Steveb-p: How would this be a BC break if this is a method (abstract) that you just extracted?

Steveb-p: Nvm, `git` trolled me and split the diff in a strange place.

:+1:

Steveb-p (approved):

lib/ResultExtractor.php
adamwojs: Fixed in https://github.com/ezsystems/ezplatform-solr-search-engine/pull/208/commits/5fcd36ec18456ca8eb55531a51d0e084b6609de2

alongosz (approved):

PR #209, opened 2021-02-10
COM-20123: Serialize bool values to solr format

wizhippo:
https://issues.ibexa.co/browse/COM-20123


wizhippo:
Just tested against solr 6.6 it it indeed will take blank for bools and set it to false where in solr 7 and 8 if you set the bool as a blank it does not add the field to the document.

wizhippo:
Solr 7+ has update processors that remove blanks

<updateProcessor class="solr.RemoveBlankFieldUpdateProcessorFactory" name="remove-blank"\/>

wizhippo:
Removal of <updateRequestProcessorChain name="add-unknown-fields-to-the-schema"> from the docker script did not match on the newer versions.  Added and all is good again adjusting the Dockerfile-solr.

PR #214, opened 2021-06-08
IBX-522: Introduced Docker image for Solr standalone

Steveb-p:
| ----------------------------------------- | ------------------
| **JIRA issue**                            | [IBX-522](https://issues.ibexa.co/browse/IBX-522)
| **Type**                                  | improvement
| **Target Ibexa DXP version**              | `v4.0`, `v3.3`
| **BC breaks**                             | no

## :exclamation: Warning :exclamation: 

**Current version of this PR builds an invalid image. Somehow files copied during build time from the extended context differ from the ones that were present when they were copied into `docker/solr` directory.**

One file has been renamed: `schema.xml` -> `managed-schema`, as that's the file that is actually loaded in the Solr docker image, apparently. `schema.xml` had no effect if left as is.


docker/solr/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM solr:8.6.3
+
+LABEL org.opencontainers.image.source=https://github.com/ezsystems/ezplatform-solr-search-engine
+
+USER root
+COPY conf server/solr/configsets/_default/conf
+RUN sed --in-place '/<updateRequestProcessorChain name="add-unknown-fields-to-the-schema".*/,/<\/updateRequestProcessorChain>/d' server/solr/configsets/_default/conf/solrconfig.xml
+RUN sed --in-place 's/${solr.autoSoftCommit.maxTime:-1}/${solr.autoSoftCommit.maxTime:20}/' server/solr/configsets/_default/conf/solrconfig.xml
```
ibexa-yuna, line 8: Why and what are those?

Steveb-p: They come from:

https://github.com/ezsystems/ezplatform-solr-search-engine/blob/90c9eb28596ec1bc7590d4009f53ddaf2b0204f8/bin/generate-solr-config.sh#L121-L123

ibexa-yuna: Are they really needed in a test environment i.e. not a long-running instance?

adamwojs (approved):

docker/solr/Dockerfile
Steveb-p: > Are they really needed in a test environment i.e. not a long-running instance?

As described by @adamwojs in a personal conversation with me, they are needed because sometimes we test against field being empty in search engine, and this apparently is required for it to work. I didn't dig deeper tbh.

adamwojs: > Are they really needed in a test environment i.e. not a long-running instance?

Yes, these config changes are required for test environment.

ViniTou:
Please take a look on:
https://github.com/ibexa/installer/pull/20

Steveb-p:
Configuration file duplication can be avoided by using project directory as `docker build` context (and passing Dockerfile location, not using default):
```shell
docker build -f docker/solr/Dockerfile .
```

PR #226, opened 2021-12-14
IBX-1699: Bump defaults and requirements for Java, Solr, Elastic

glye:
Install Solr 8.11.1 by default, as this includes the upgraded log4j with an important security fix.
For 1.7 branch, eZ Platform v2.5.
Maybe best to release as v1.8.0 with a new v2.5 release, could be compatibility issues.

Work in progress:
- [x] Install Solr 8.11.1 - ~this will fail as 8.11.1 is not available yet at https://archive.apache.org/dist/lucene/solr/~ (available now)
- [ ] Readme update is incomplete
- [ ] Nothing for Elasticsearch and Java yet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,19 +11,29 @@ cache:
 env:
     global:
         - COMPOSER_MEMORY_LIMIT=6G
+        - TEST_CONFIG="phpunit-integration-legacy-solr.xml"
 
 matrix:
     include:
         - php: 7.2
           env: TEST_CONFIG="phpunit.xml"
         - php: 7.1
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="6.4.2"  CORES_SETUP="dedicated"
+          env: SOLR_VERSION="7.7.3"  CORES_SETUP="dedicated"
         - php: 7.2
-          env: TEST_CONFIG="phpunit-integration-legacy-solr.xml" SOLR_VERSION="6.5.1"  CORES_SETUP="shared"
+          env: SOLR_VERSION="7.7.3"  CORES_SETUP="shared"
         - php: 7.2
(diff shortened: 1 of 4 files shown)


mnocon:
Does it make sense to update CI to test on Solr 8? It uses Solr 6 right now:
https://github.com/ezsystems/ezplatform-solr-search-engine/blob/1.7/.travis.yml#L20-L26

glye:
Right, I/we should bump that too. Given the doc we should skip 6 entirely and only test 7 and 8. I'm on vacation now - can you verify the rest first, or add the travis setting yourself? I guess we should test the 8.11.1 and the latest 7.x

mnocon:
Sure, I will push the CI changes here

mnocon:
I've looked into this and I don't think we can bump Solr to 7.7 in 1.7 branch.

Please have a look at https://github.com/ezsystems/ezplatform-solr-search-engine/pull/136

```
Due to the major changes in Solr, support for 7.X will be part of the ezplatform-solr-search-engine 2.0 release which will be compatible with eZ Platform 2.5 release. Solr 6.X will be dropped in this release.

Solr 7.X and Solr 8.X will be officialy supported in ezplatform-solr-search-engine 3.0 release which will be compatible with eZ Platform 3.X.
```

IMHO we should consider bumping the package version used by 2.5 (product) to 2.0 by default - it's not update-friendly, but with good upgrade doc people needing to stay on Solr 6 will be able to manually downgrade to 1.7 and others will be secure by default.

glye:
Right, thanks. In that case we should drop this PR, and instead do a PR to have platform 2.5 bumped to 2.0 of this package. And we should see if we need upgrades in the 2.0 and 3.x branches, for the most recent solr versions.

glye:
Upgrade Platform v2.5 to v2 of this repo instead: https://github.com/ezsystems/ezplatform/pull/677
May need to do a PR against v2 of this repo to bump Solr.

glye:
Added PR for v2 with your changes, @mnocon 
https://github.com/ezsystems/ezplatform-solr-search-engine/pull/228

PR #235, opened 2022-05-19
IBX-2618: Optimized bulk indexing

alongosz:
| ---------------------------------------- | ------------------
| **JIRA issue**                          | [IBX-2618](https://issues.ibexa.co/browse/IBX-2618)
| **Requires**                           | ezsystems/ezplatform-kernel#306
| **Type**                                   | improvement
| **Target Ibexa version** | `v3.3`+
| **BC breaks**                          | no
| **Tests**                              | Passing [integration build](https://github.com/ezsystems/ezplatform-solr-search-engine/actions/runs/2448028950), requires ezsystems/ezplatform-kernel#306

This PR provides a series of optimizations in order to improve search indexing and also fixes minor issues

- [x] **Optimized calling the same Endpoint multiple times**
       Documents were bulk-grouped by a language instead of an Endpoint. While in the case of endpoint per language setups it works, it's not very optimal when there's one endpoint handling multiple languages. Changed this so Documents are grouped per Endpoint.
- [x] **Replaced magic getters with strict ones in Search FieldValueMappers**
      See ezsystems/ezplatform-kernel#306 for more details.
- [x] **Split UpdateSerializer into XML and JSON format strategies**
      This allowed us to change internally indexing format to JSON, which is lighter. It's possible to go back to XML via container parameter. Not intended as an official configuration, but rather "just in case".
- [x] **Refactored legacy HttpClient to rely on symfony/http-client**
       Seems like legacy HttpClient for quite some time reinvented the wheel. Symfony HttpClient is more versatile, handles more cases, including compression if both sides support it.
- [x] **Fixed mapping of content_language_codes Solr fields**
       Mapping should be done based on `VersionInfo::$languageCodes` rather than keys of loaded names array
- [x] **Logged failed Solr HTTP connection attempts**
       Done so we can have more idea of the state of the infrastructure while debugging long duration of network requests.

### TODO:
- [x] DROP TMP COMMIT

#### Checklist:
- [x] Provided PR description.
- [x] Tested the solution manually.
- [x] Provided automated test coverage.
- [x] Checked that target branch is set correctly
- [x] Ran PHP CS Fixer for new PHP code (use `$ composer fix-cs`).
- [x] Asked for a review

### DIFF
--- a/lib/Gateway/HttpClient/Stream.php
+++ b/lib/Gateway/HttpClient/Stream.php
@@ -12,7 +12,7 @@ use EzSystems\EzPlatformSolrSearchEngine\Gateway\Message;
 use Psr\Log\LoggerAwareTrait;
 use Psr\Log\LoggerAwareInterface;
 use Psr\Log\NullLogger;
-use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\ExceptionInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 /**
@@ -27,32 +27,16 @@ class Stream implements HttpClient, LoggerAwareInterface
     /** @var int */
     private $connectionTimeout;
 
-    /** @var int */
-    private $connectionRetry;
-
-    /** @var int */
-    private $retryWaitMs;
-
     /** @var \Symfony\Contracts\HttpClient\HttpClientInterface */
     private $client;
 
     /**
-     * Stream constructor.
-     *
      * @param int $timeout Timeout for connection in seconds.
-     * @param int $retry Number of times to re-try connection.
-     * @param int $retryWaitMs Time in milliseconds.
      */
-    public function __construct(
-        HttpClientInterface $client,
-        int $timeout = 10,
-        int $retry = 5,
-        int $retryWaitMs = 100
-    ) {
+    public function __construct(HttpClientInterface $client, int $timeout = 10)
+    {
         $this->client = $client;
         $this->connectionTimeout = $timeout;
-        $this->connectionRetry = $retry;
-        $this->retryWaitMs = $retryWaitMs;
         $this->setLogger(new NullLogger());
     }
 
@@ -64,51 +48,22 @@ class Stream implements HttpClient, LoggerAwareInterface
      * @param string $method
      * @param string $path
      *
-     * @throws \Symfony\Contracts\HttpClient\Exception\ServerExceptionInterface
-     * @throws \Symfony\Contracts\HttpClient\Exception\ClientExceptionInterface
-     * @throws \Symfony\Contracts\HttpClient\Exception\RedirectionExceptionInterface
+     * @throws \EzSystems\EzPlatformSolrSearchEngine\Gateway\HttpClient\ConnectionException
      */
     public function request($method, Endpoint $endpoint, $path, Message $message = null): Message
     {
         $message = $message ?? new Message();
 
-        // We'll try to reach backend several times before throwing exception.
-        $i = 0;
-        do {
-            ++$i;
-            try {
-                if (
-                    $responseMessage = $this->getResponseMessage(
-                        $method,
-                        $endpoint,
-                        $path,
-                        $message
-                    )
-                ) {
-                    return $responseMessage;
-                }
-            } catch (TransportExceptionInterface $e) {
-                $this->logger->warning(
-                    sprintf(
-                        'Connection attempt #%d to %s failed, retrying after %d ms',
-                        $i,
-                        $endpoint->getURL(),
-                        $this->retryWaitMs
-                    )
-                );
-            }
-            usleep($this->retryWaitMs * 1000);
-        } while ($i < $this->connectionRetry);
-
-        $this->logger->error(
-            sprintf(
-                'Connection to %s failed, attempted %d times',
-                $endpoint->getURL(),
-                $this->connectionRetry
-            )
-        );
-
-        throw new ConnectionException($endpoint->getURL(), $path, $method);
+        try {
+            return $this->getResponseMessage(
+                $method,
+                $endpoint,
+                $path,
+                $message
+            );
+        } catch (ExceptionInterface $e) {
+            throw new ConnectionException($endpoint->getURL(), $path, $method, $e);
+        }
     }
 
     /**
--- a/lib/Gateway/HttpClient/Stream.php
+++ b/lib/Gateway/HttpClient/Stream.php
@@ -12,6 +12,8 @@ use EzSystems\EzPlatformSolrSearchEngine\Gateway\Message;
 use Psr\Log\LoggerAwareTrait;
 use Psr\Log\LoggerAwareInterface;
 use Psr\Log\NullLogger;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 /**
  * Simple PHP stream based HTTP client.
@@ -22,20 +24,17 @@ class Stream implements HttpClient, LoggerAwareInterface
 {
     use LoggerAwareTrait;
 
-    /**
-     * @var int
-     */
+    /** @var int */
     private $connectionTimeout;
 
-    /**
-     * @var int
-     */
+    /** @var int */
     private $connectionRetry;
 
-    /**
-     * @var int
-     */
+    /** @var int */
     private $retryWaitMs;
+
+    /** @var \Symfony\Contracts\HttpClient\HttpClientInterface */
+    private $client;
 
     /**
      * Stream constructor.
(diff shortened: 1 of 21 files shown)


lib/Gateway/HttpClient/Stream.php
```diff
@@ -1,178 +1,154 @@
...
+                    return $responseMessage;
+                }
+            } catch (TransportExceptionInterface $e) {
+                $this->logger->warning(
+                    sprintf(
+                        'Connection attempt #%d to %s failed, retrying after %d ms',
+                        $i,
+                        $endpoint->getURL(),
+                        $this->retryWaitMs
+                    )
```
Steveb-p: Pass exception to logger context.

```suggestion
                $this->logger->warning(
                    sprintf(
                        'Connection attempt #%d to %s failed, retrying after %d ms',
                        $i,
                        $endpoint->getURL(),
                        $this->retryWaitMs
                    ), [
                        'exception' => $e,
                    ]
```

lib/Gateway/HttpClient/Stream.php
```diff
@@ -1,178 +1,154 @@
...
-                list($key, $value) = explode(':', $lineContent, 2);
-                $headers[$key] = ltrim($value);
-            }
-        }
-
-        return new Message(
-            $headers,
-            $body
-        );
+        throw new ConnectionException($endpoint->getURL(), $path, $method);
```
Steveb-p: If this exception is not specifically handled, then callling logger above would make it so it becomes logged twice (here & by normal application exception handler).

src/lib/Gateway/UpdateSerializerFactory.php
```diff
@@ -0,0 +1,42 @@
...
+
+class UpdateSerializerFactory
+{
+    /** @var \Ibexa\Solr\Gateway\UpdateSerializerInterface[]|iterable */
+    private $serializers;
+
+    /**
+     * @param array<\Ibexa\Solr\Gateway\UpdateSerializerInterface> $serializers
+     */
+    public function __construct(iterable $serializers)
```
Steveb-p: This may confuse the static analysis and hide an error when someone tries to access this like an array, when it may not be one.

Side note:
Additionally, you can make use of `index_by` functionality of `!tagged_iterator` to remove the need for serializers to have `getSupportedFormat()` method. This would also make it possible for a serializer to be registered more than once, for multiple formats.

src/lib/Gateway/UpdateSerializerFactory.php
```diff
@@ -0,0 +1,42 @@
...
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\Solr\Gateway;
+
+use eZ\Publish\SPI\Exception\InvalidArgumentException;
+
+class UpdateSerializerFactory
```
Steveb-p, line 13: It's not a factory, it's a registry, which accidentally performs a role of factory during container compilation.

(which, btw, _almost_ makes it a circular reference)

damianz5 (approved):

lib/Gateway/HttpClient/Stream.php
```diff
@@ -1,178 +1,154 @@
...
-            if (is_numeric($name)) {
-                throw new \RuntimeException("Invalid HTTP header name $name");
-            }
+        $response = $this->client->request(
+            $method,
+            $endpoint->getURL() . $path,
+            [
+                'headers' => $message->headers,
+                'timeout' => $this->connectionTimeout,
+                'body' => $message->body,
```
damianz5, line 91: consider adding logger->debug call for body size

alongosz: > consider adding logger->debug call for body size

The point of all those changes were to increase performance. Don't you think that adding extra I/O, in the form of logging, for a normal expected flow, decreases that performance?

Steveb-p: These things can be already achieved by setting logger onto the http client itself.

lib/Gateway/HttpClient/Stream.php
```diff
@@ -1,178 +1,154 @@
...
+                }
+            } catch (TransportExceptionInterface $e) {
+                $this->logger->warning(
+                    sprintf(
+                        'Connection attempt #%d to %s failed, retrying after %d ms',
+                        $i,
+                        $endpoint->getURL(),
+                        $this->retryWaitMs
+                    )
+                );
```
Steveb-p: Retry functionality can be handled internally by http client:

https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/HttpClient/RetryableHttpClient.php

tests/bundle/Gateway/UpdateSerializer/JsonUpdateSerializerTest.php
```diff
@@ -0,0 +1,166 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformSolrSearchEngineBundle\Tests\Gateway\UpdateSerializer;
```
adamwojs: This should be placed under Ibexa\ namespace

tests/lib/Gateway/UpdateSerializerFactoryTest.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformSolrSearchEngine\Tests\Gateway;
```
adamwojs: This should be placed under Ibexa\ namespace

lib/Gateway/HttpClient/Stream.php
damianz5: Hard to tell without comparison tests, probably insignificant change for NullLogger or configuration higher than debug level. As of performance - using NullLogger could increase it more [ref](https://github.com/symfony/symfony/pull/14682)

Steveb-p: `NullLogger` could negatively impact performance if there is a lot of data transformation done when posting log message onto Logger itself (which should not happen, it's log message processors job to do that).

In this case however, since we're working with things that are already in memory, I would say it's mostly irrelevant.

Despite this, I'd say we should stop considering it at this point in time and instead focus our attention elsewhere. When client gets decorated with retryable client & provided with correct loggers this issue should disappear on it's own, as retries & data logging will be done using built-in functionality. Let's NOT add debugging info here and prepare to close the issue. Code introducing decorated variants can be done as a follow up for this.

alongosz:
Update: via 4f7ef6f delegated the whole connection re-try procedure to `symfony/http-client`.

lib/Gateway/HttpClient/Stream.php
alongosz: Done via 4f7ef6f

src/lib/Gateway/UpdateSerializerFactory.php
alongosz: > This may confuse the static analysis and hide an error when someone tries to access this like an array, when it may not be one.

So what should be done here instead?

> 
> Side note: Additionally, you can make use of `index_by` functionality of `!tagged_iterator` to remove the need for serializers to have `getSupportedFormat()` method. This would also make it possible for a serializer to be registered more than once, for multiple formats.

This is done on purpose, it's not used just by the factory:
>  `'Content-Type' => 'application/' . $this->updateSerializer->getSupportedFormat(),`

Serializer's responsibility is supporting one format compatible with Solr Content-Type. It's not within the scope of DI.

src/lib/Gateway/UpdateSerializerFactory.php
alongosz: > It's not a factory, it's a registry, which accidentally performs a role of factory during container compilation.
> 
> (which, btw, _almost_ makes it a circular reference)

Don't you think it would be confusing to call it a Registry when it's actually used one time during compilation, like this:
``` yaml
Ibexa\Solr\Gateway\UpdateSerializerInterface:
        factory: ['@Ibexa\Solr\Gateway\UpdateSerializerFactory', 'getSerializer']
```
?

tests/lib/Gateway/UpdateSerializerFactoryTest.php
alongosz: > This should be placed under Ibexa\ namespace

Fixed both namespaces via 41a1413, though I did that rather on purpose, to avoid composer.json hassle. Now we have 4 namespaces and 2 of them (legacy ones, depends on the order) are reported by PHPStorm as incorrect, which causes a lot of false negative issues. There's no BC really needed nor possible for `autoload-dev` namespaces, so I left it for upper branch.

Nevertheless according to our own agreed new code policy - changed as requested.

src/lib/Gateway/UpdateSerializerFactory.php
alongosz: > > This may confuse the static analysis and hide an error when someone tries to access this like an array, when it may not be one.
> 
> So what should be done here instead?
> 

Fixed via bcff6e7

src/lib/Gateway/UpdateSerializerFactory.php
Steveb-p: If that's the only reason for it's existence, then by the same token we could simply alias the service and replace the format parameter with service ID (which would be, in this case, it's class name).

From my point of view it's a registry:
 * It stores services
 * You can get a specific service by using it's method
 * Serializer in question exists without this service
 * It doesn't construct any services on it's own

You can easily imagine a `UpdateSerializerFactory::getSerializer($format)` instead. The fact that we default to a specific format (and don't provide an argument) is a detail.

alongosz: > If that's the only reason for it's existence, then by the same token we could simply alias the service and replace the format parameter with service ID (which would be, in this case, it's class name).
> 
> From my point of view it's a registry:
> 
> * It stores services
> * You can get a specific service by using it's method
> * Serializer in question exists without this service
> * It doesn't construct any services on it's own
> 
> You can easily imagine a `UpdateSerializerFactory::getSerializer($format)` instead. The fact that we default to a specific format (and don't provide an argument) is a detail.

We have 2 serializers. While we don't plan for anyone to change it back to XML, currently it's quite simple - via service container parameter. With the change you're proposing the only way to do it would be via decoration or override, which, when it comes to extensibility points is a bit less stable.

@Steveb-p Do you really feel it brings more value to simplify it to reduce maintenance cost while getting rid of flexibility?

Steveb-p (approved):

src/lib/Gateway/UpdateSerializerFactory.php
Steveb-p: What I'm primarily focused on is that it is **not**, in fact, a factory. It does not create objects, it merely stores them.

tests/lib/Gateway/UpdateSerializerFactoryTest.php
ViniTou: composer.json shenanigans are much easier to handle then troubles with merging things up. :D

konradoboza (approved):

micszo (approved):
Test status https://github.com/ezsystems/ezplatform-kernel/pull/306#pullrequestreview-1018138908.

PR #239, opened 2022-07-12
IBX-3116: Made Section identifier nonconvertible

barw4:
| ----------------------------------------- | ------------------
| **JIRA issue**                            | [IBX-3116](https://issues.ibexa.co/browse/IBX-3116)
| **Type**                                  | bug
| **Target Ibexa DXP version**              | `v3.3`
| **BC breaks**                             | no

As the title suggests.

### DIFF
--- a/lib/FieldMapper/ContentFieldMapper/BlockDocumentsBaseContentFields.php
+++ b/lib/FieldMapper/ContentFieldMapper/BlockDocumentsBaseContentFields.php
@@ -166,7 +166,7 @@ class BlockDocumentsBaseContentFields extends ContentFieldMapper
             new Field(
                 'content_section_identifier',
                 $section->identifier,
-                new FieldType\StringField()
+                new FieldType\IdentifierField(['raw' => true])
             ),
             new Field(
                 'content_section_name',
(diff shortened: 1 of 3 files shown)


adamwojs (changes requested):
Are you sure you know what you are doing?

Instead of changing type to string please use raw flag on identifier field type or change string normalization in a way to that underscores will be accepted. 

Please note that similar changes need to be applied for elasticsearch integration and integration test is mandatory in this case.

barw4:
@adamwojs yes, I'm sure. I was thinking about changing converter to accept "-", but that drastically changes schema in other places.

Besides, we do have strings here as well: https://github.com/ezsystems/ezplatform-solr-search-engine/blob/9d665367d9e6599b395efb721ea3ddbbdc91cef5/lib/FieldMapper/ContentFieldMapper/BlockDocumentsBaseContentFields.php#L187-L190 - those are also identifiers?

adamwojs (approved):

Steveb-p (approved):
Please be aware that in cases like this describing what the PR contents are does not describe **why** are they necessary.

bogusez (approved):

adamwojs:
Could you please merge up changes?

PR #241, opened 2022-09-19
IBX-3863: Parametrized Solr HTTP Client timeout and max retries

alongosz:
| Question                                  | Answer |
| ---------------------------------------- | ------------------ |
| **JIRA issue**                          | [IBX-3863](https://issues.ibexa.co/browse/IBX-3863) |
| **Type**                                   | improvement |
| **Target Ibexa version** | `v3.3`+ |
| **BC breaks**                          | no |

When switching to Symfony HTTP Client to handle Solr queries in #235. I didn't introduce any possibility to configure Client's timeout nor number of max retries. There are some edge cases which show that this parameter might work-around performance issues.

### Documentation

~Solr~ Ibexa Solr Bundle uses Symfony HTTP Client to fetch and update Solr index. It's possible to configure timeout and number of max retries for that client using Solr Bundle's Semantic Configuration:

v3.3:
```yaml
ez_search_engine_solr:
    http_client:
        timeout: 30
        max_retries: 5
```

v4.x:
```yaml
ibexa_solr:
    http_client:
        timeout: 30
        max_retries: 5
```

### QA

You can test timeout change by mocking Solr service
```
$ nc -vv -l 8983
```
and then trying to perform e.g. reindexing
```
php ./bin/console ibexa:reindex --processes 1 -n
```

See that modifying both `timeout` and `max_retries` affect the wait time and the amount of tries before failure.

## Open questions

Given this rather should be avoided, preferring proper performance fixes, should we:
- [x] Document those parameters?
- [x] Provide Semantic configuration instead of parameters (only if we decide to document)

#### Checklist:
- [x] Provided PR description.
- [x] Tested the solution manually.
- [x] Provided automated test coverage.
- [x] Checked that target branch is set correctly.
- [x] Asked for a review

### DIFF
--- a/bundle/DependencyInjection/EzSystemsEzPlatformSolrSearchEngineExtension.php
+++ b/bundle/DependencyInjection/EzSystemsEzPlatformSolrSearchEngineExtension.php
@@ -20,7 +20,7 @@ use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
 use Symfony\Component\Config\FileLocator;
 
 /**
- * @psalm-type SolrHttpClientConfigArray = array{timeout: int, max_retries: int}
+ * @phpstan-type SolrHttpClientConfigArray = array{timeout: int, max_retries: int}
  */
 class EzSystemsEzPlatformSolrSearchEngineExtension extends Extension
 {
@@ -324,7 +324,7 @@ class EzSystemsEzPlatformSolrSearchEngineExtension extends Extension
     }
 
     /**
-     * @psalm-param SolrHttpClientConfigArray $httpClientConfig
+     * @phpstan-param SolrHttpClientConfigArray $httpClientConfig
      */
     private function configureHttpClient(ContainerBuilder $container, array $httpClientConfig): void
     {
--- a/bundle/DependencyInjection/EzSystemsEzPlatformSolrSearchEngineExtension.php
+++ b/bundle/DependencyInjection/EzSystemsEzPlatformSolrSearchEngineExtension.php
@@ -19,6 +19,9 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
 use Symfony\Component\Config\FileLocator;
 
+/**
+ * @psalm-type SolrHttpClientConfigArray = array{timeout: int, max_retries: int}
+ */
 class EzSystemsEzPlatformSolrSearchEngineExtension extends Extension
 {
     /**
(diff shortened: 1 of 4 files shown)


Steveb-p (approved):
Semantic bundle configuration is almost always preferred and recommended by Symfony.

konradoboza (approved):

ciastektk (approved):

Steveb-p (approved):
What you can additionally do is to pass additional options to `ibexa.solr.http_client` factory.

What I mean specifically, we can pass `extra.curl.CURLOPT_CONNECTTIMEOUT` (`extra` contains client-specific configurations) so that if we are having trouble establishing a connection, then we're dropping the attempt earlier (like after 2 seconds maybe?).

This will help us differentiate between service being overloaded / responding longer vs. service being completely unavailable.

See https://symfony.com/doc/current/http_client.html#configuring-curlhttpclient-options

alongosz:
> What you can additionally do is to pass additional options to `ibexa.solr.http_client` factory.
> 
> What I mean specifically, we can pass `extra.curl.CURLOPT_CONNECTTIMEOUT` (`extra` contains client-specific configurations) so that if we are having trouble establishing a connection, then we're dropping the attempt earlier (like after 2 seconds maybe?).
> 
> This will help us differentiate between service being overloaded / responding longer vs. service being completely unavailable.
> 
> See https://symfony.com/doc/current/http_client.html#configuring-curlhttpclient-options

@Steveb-p I got the error when trying to set that option

```
Cannot set "CURLOPT_CONNECTTIMEOUT" with "extra.curl"
```

Moreover I think it will be the opposite of what we're trying to achieve. We need to extend connecting time in case of slow solr instance, but we want to avoid retries when instance is not running at all.

ViniTou (approved):

konradoboza (approved):

bundle/DependencyInjection/EzSystemsEzPlatformSolrSearchEngineExtension.php
```diff
@@ -19,6 +19,9 @@
 use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
 use Symfony\Component\Config\FileLocator;
 
+/**
+ * @psalm-type SolrHttpClientConfigArray = array{timeout: int, max_retries: int}
```
Steveb-p: `@psalm-*` notation is understood by phpstan (and probably the other way around too), but I think we should stick to one we already selected as our static analysis, and that is phpstan.

Steveb-p (approved):

tomaszszopinski (approved):
QA approved on IbexaDXP 3.3 commerce.

PR #243, opened 2023-02-20
IBX-5133: Allowed Solr tests to be executed against different hostname

Steveb-p:
https://issues.ibexa.co/browse/IBX-5133

This PR allows usage of `SOLR_HOST` environment variable to control the actual hostname of the Solr instance.

`env(SOLR_HOST)` notation in the service declaration files was attempted, but failed. Apparently our custom service container does not properly resolve environment variables in it's expressions, so I have dropped the idea of trying to fix it after a few attempts.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,8 +62,6 @@
         }
     },
     "config": {
-        "allow-plugins": {
-            "*": false
-        }
+        "allow-plugins": false
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,7 @@
     },
     "config": {
         "allow-plugins": {
-            "php-http/discovery": false
+            "*": false
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -60,5 +60,10 @@
         "branch-alias": {
             "dev-master": "3.3.x-dev"
         }
+    },
+    "config": {
+        "allow-plugins": {
+            "php-http/discovery": false
+        }
     }
 }
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -6,6 +6,7 @@
          convertWarningsToExceptions="true">
     <php>
         <ini name="error_reporting" value="-1" />
+        <env name="SOLR_HOST" value="localhost" />
     </php>
     <testsuites>
         <testsuite name="Solr search engine tests">
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -6,7 +6,6 @@
          convertWarningsToExceptions="true">
     <php>
         <ini name="error_reporting" value="-1" />
-        <env name="SOLR_HOST" value="localhost" />
     </php>
     <testsuites>
         <testsuite name="Solr search engine tests">
--- a/tests/lib/Resources/config/cloud.yml
+++ b/tests/lib/Resources/config/cloud.yml
@@ -33,7 +33,7 @@ services:
         arguments:
             -
                 scheme: http
-                host: localhost
+                host: "%test.ibexa.solr.host%"
                 port: 8983
                 path: /solr
                 core: core0
@@ -45,7 +45,7 @@ services:
         arguments:
             -
                 scheme: http
-                host: localhost
+                host: "%test.ibexa.solr.host%"
                 port: 8983
                 path: /solr
                 core: core1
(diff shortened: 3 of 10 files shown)


composer.json
```diff
@@ -60,5 +60,10 @@
         "branch-alias": {
             "dev-master": "3.3.x-dev"
         }
+    },
+    "config": {
+        "allow-plugins": {
+            "*": false
+        }
     }
```
alongosz:
```suggestion
    "config": {
        "allow-plugins": false
    }
```

phpunit.xml
```diff
@@ -6,6 +6,7 @@
          convertWarningsToExceptions="true">
     <php>
         <ini name="error_reporting" value="-1" />
+        <env name="SOLR_HOST" value="localhost" />
```
alongosz: this should rather be needed for integration tests only, right?

tests/lib/Resources/config/cloud.yml
```diff
@@ -33,7 +33,7 @@ services:
         arguments:
             -
                 scheme: http
-                host: localhost
+                host: "%test.ibexa.solr.host%"
```
alongosz: CS
```suggestion
                host: '%test.ibexa.solr.host%'
```

tests/lib/Resources/config/cloud.yml
```diff
@@ -45,7 +45,7 @@ services:
...
```
alongosz: same CS comment here and later in similar places.

tests/lib/SetupFactory/LegacySetupFactory.php
```diff
@@ -63,6 +63,8 @@ public function getRepository($initializeFromScratch = true)
      */
     protected function externalBuildContainer(ContainerBuilder $containerBuilder): void
     {
+        $containerBuilder->setParameter('test.ibexa.solr.host', getenv('SOLR_HOST') ?: 'localhost');
```
alongosz, line 66: Hmm, I remember issues with getenv and PHPUnit. Was it that what is being set in phpunit.xml doesn't get set in `$_ENV`?

Steveb-p: It most certainly is, as I was testing it locally and switching to a different hostname in this very fashion.

phpunit.xml
Steveb-p: Yes, but it is way more visible this way. Otherwise you would need to specifically look into code, into service definitions, to see that an environment variable is available (not mentioning readme, but who reads those anyway...).

EDIT: And if you're asking to split up integration & unit tests, that's fine - but a little bit out of scope. It may require that little bit of precious time that we currently do not have :sweat_smile:

tests/lib/Resources/config/cloud.yml
Steveb-p: It follows the style of all other quotations in this very file. Do you want me to update those immediately, only these, or leave as-is?

alongosz: > It follows the style of all other quotations in this very file. Do you want me to update those immediately, only these, or leave as-is?

We can update it step by step. We've agreed not to use old CS in new code. Historically this was my mistake, I thought for some reason that these are getting parsed and updated big portion of the codebase to use double quotes. This wasn't correct, pre-processing doesn't happen here, so it's a waste to put double quotes. Long time ago after that we've decided with the Team that all Yaml literals will be single quoted, and quoted only if needed (so mainly service and parameters injection, and colons inside of literals)

phpunit.xml
alongosz: > Yes, but it is way more visible this way. Otherwise you would need to specifically look into code, into service definitions, to see that an environment variable is available (not mentioning readme, but who reads those anyway...).
> 
> EDIT: And if you're asking to split up integration & unit tests, that's fine - but a little bit out of scope. It may require that little bit of precious time that we currently do not have sweat_smile

I think the issue here is that Solr integration tests should be launched using `vendor/ezsystems/ezplatform-kernel/phpunit-integration-legacy-solr.xml` config.

I'm not saying it's good, all I'm saying it was like that by design, which CI follows. So I still have doubts about it. Need to take closer look.

Ref.: https://github.com/ezsystems/ezplatform-solr-search-engine/actions/runs/4223675904/jobs/7333688945#step:8:12

kisztof (approved):

konradoboza (approved):

PR #245, opened 2023-07-24
IBX-6221: removed duplicate conditions from isFieldEmpty visitor

reithor:
https://issues.ibexa.co/browse/IBX-6221

Removes duplicate sql conditions from `FieldEmpty->visit`
(does not affect query result but shortens query that is send to solr)

### DIFF
--- a/lib/Query/Common/CriterionVisitor/Field/FieldEmpty.php
+++ b/lib/Query/Common/CriterionVisitor/Field/FieldEmpty.php
@@ -77,6 +77,6 @@ final class FieldEmpty extends Field
             }
         }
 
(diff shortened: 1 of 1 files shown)


alongosz (commented):
Looks good.

We need some basic unit test coverage for that use case, specifically for that `FieldEmpty` Visitor, showcasing the bug. The test should create a `Criterion` which caused duplication and check if the output is as expected.

So, expecting to see new `tests/Query/Common/CriterionVisitor/Field/FieldEmptyTest.php` :)
Please cover just the essence of the bug - I know we could do much more there but that's on us - maintainers :)

PR #250, opened 2024-06-10
IBX-8378: Fixed handling non-indexable field types

vidarl:
| :ticket: Issue | [IBX-8378](https://issues.ibexa.co/browse/IBX-8378) |
|----------------|-----------|

When making a custom field type, it must implement [Indexable](https://doc.ibexa.co/en/latest/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-FieldType-Indexable.html#methods) (even though the field content should **not** be searchable) or else the solr bundle will return exception:
```
OutOfBoundsException:
Field Type "mycustomfieldtype" is not indexable. Provide Ibexa\Contracts\Core\FieldType\Indexable implementation and register it with the "ibexa.field_type.indexable" tag.
 
  at vendor/ibexa/core/src/lib/Search/Common/FieldRegistry.php:39
  at Ibexa\Core\Search\Common\FieldRegistry->getType('ezuser')
     (vendor/ibexa/solr/src/lib/FieldMapper/ContentTranslationFieldMapper/BlockDocumentsContentFields.php:79)
```

Legacy search and Elastic search will not throw any errors.

It is a bit unclear to me if it is intended that all fieldtypes must implement Indexable or not.
pro:
 - All core fieldtypes implement Indexable, [also those that is not searchable](https://github.com/ezsystems/ezplatform-kernel/blob/c88c39759cb12c701314e541a8ae499e6f70bfca/eZ/Publish/Core/settings/indexable_fieldtypes.yml#L119-L138)

cons:
 - [Documentation](https://doc.ibexa.co/en/latest/content_management/field_types/field_type_search/) doesn't say it is required unless you want data be searchable
 - All other search engine adapters (legacy search and elastic) doesn't require it.

So either documentation should be updated or Solr bundle should behave as the other search engine adapters. This PR adds support in Solr for not having to implement Indexable


lib/FieldMapper/ContentTranslationFieldMapper/BlockDocumentsContentFields.php
```diff
@@ -74,7 +74,7 @@ public function mapFields(Content $content, $languageCode)
             }
 
             foreach ($contentType->fieldDefinitions as $fieldDefinition) {
-                if ($fieldDefinition->id !== $field->fieldDefinitionId) {
+                if ($fieldDefinition->id !== $field->fieldDefinitionId || !$fieldDefinition->isSearchable) {
```
alongosz, line 77: AFAICS indeed Indexable can be skipped within the current Repository's Search architecture. What I don't follow is the fact that you're checking here `isSearchable` instead of checking if there's indexable implementation (not sure if this can be directly inferred from the variables you have here though).

Q:

1. Does not implementing `Indexable` mean that `isSearchable` is false?
2. Am I able now to uncheck Is Searchable for a field type that is indexable?
3. Is isSearchable unchecked for fields that are not Indexable? (that probably would answer my initial comment above.

vidarl: This is basically how [ElasticSearch implementation does the same thing](https://github.com/ezsystems/ezplatform-elastic-search-engine/blob/1.2/src/lib/DocumentMapper/DocumentFactory.php#L337).

1. Does not implementing `Indexable` mean that `isSearchable` is false?
1.1 Yes. In admin-ui "Searchable" is grayed out. On API level setting FieldDefinitionCreateStruct.isSearchable=true will cause validation error
2. Am I able now to uncheck Is Searchable for a field type that is indexable?
2.1 Yes, as before you need to re-index in order to make it have any effect on existing content though.
3. Is isSearchable unchecked for fields that are not Indexable? (that probably would answer my initial comment above.
3.1 Yes, it is

So the easiest way is to simply check for `isSearchable`. It could be an option to add a [isIndexable member to FieldRegistry](https://github.com/ezsystems/ezplatform-kernel/blob/1.3/eZ/Publish/Core/Search/Common/FieldRegistry.php#L15). But in the mapper you would then ending up having to check for both isSearchable and isIndexable so it will only have a semantic ( it that is the correct word) meaning

alongosz: Thanks for the clarification :+1:

alongosz (approved):

konradoboza (approved):

tomaszszopinski (approved):
QA approved on IbexaDXP 3.3 commerce.

